PR #11, opened 2016-09-20
Allow payment with an amount set to 0

rchomat:


rikterbeek:
Hi @rchomat,

Thanks for the PR. The code that your referring to already allows a "0" input.
I think even better is just to use if(!$params[$parent][$child]) { ...}

Regards,
Rik
Adyen

rchomat:
Hi @rikterbeek 

Thank you for the answer.
The goal of this PR is to allow a payment set to 0 in order we can update a customer card for a reccuring contract without charge the account. Indeed, our customer may need to update their card info for a future payment.

The code you just give me doesn't change anything. The result would be the same and I could not do an update of card without charge the card.
```php
$params[$parent][$child] = 0;
var_dump(!$params[$parent][$child]); // returns true because if different than false
```

rikterbeek:
Hi @rchomat,

Thanks for the clarifying. For now just use "0" instead of 0. I tested your PR and will merge it into our library.

Regards,
Rik
Adyen

rikterbeek:
Hi @rchomat,

You did the PR to the master could you please PR this to the develop branch.

Regards,
Rik
Adyen

rchomat:
Hi @rikterbeek,

Here is the PR on develop branch : https://github.com/Adyen/adyen-php-api-library/pull/13

Regards,
Rémi

Issue #12, opened 2016-09-25
[PW-1066] Use PSR for HTTP client


jorge07:
It's possible to use Guzzle to use PSR standard and have better integrations? Or php http to be agnostic...

rikterbeek:
Hi @jorge07,

What do you mean by better integrations ?
For now we have chosen to use a custom Curl integration to keep the dependencies as minimum as possible. 

Regards,
Rik
Adyen

jorge07:
Hi @rikterbeek 

I think it's easier deal with PSR-7 rather than a custom responses as simple arrays. Also can be difficult to integrate with other libs due to this non standard usage, and there is no options for middlewares, or a way to prepare a request to be sent... Packages like https://github.com/php-http/httplug gives you an abstraction to let the final user implement his own PSR-7 http client via DI, or you can go directly to guzzle... An standard its always better.

Another question, do you accept pull requests?

rikterbeek:
Hi @jorge07,

We will look into your suggestion. Yes we do!

Regards,
Rik
Adyen

PR #24, opened 2017-05-03
Made error handlers protected instead of private

ivodvb:
I wanted to extend the CurlClient of Adyen to add some custom functionality to the HTTP client. 
I created my own CurlClient class which extends Adyens' CurlClient. I only wanted to add some small pieces of code, so I want to reuse as much as possible. Currently the methods handleCurlError and handleResultError are private. 
I want to suggest to make them protected so they can be inherited.

Issue #12 would be even better, but hasn't been resolved yet and I don't have enough time to do that.

### DIFF
--- a/src/Adyen/HttpClient/CurlClient.php
+++ b/src/Adyen/HttpClient/CurlClient.php
@@ -184,7 +184,7 @@ class CurlClient implements ClientInterface
      * @param $logger
      * @throws \Adyen\AdyenException
      */
-    private function handleCurlError($url, $errno, $message, $logger)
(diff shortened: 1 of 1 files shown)


rikterbeek:
Hi @ivodvb,

Thanks for the PR. Indeed you should be able to overwrite the HTTPClient.

Regards,
Rik
Adyen

ivodvb:
Hi @rikterbeek,

Thanks for merging. Do you have an indication for when the next release of this package will be published?

Best,
Ivo

PR #25, opened 2017-05-03
Also testing php 7.0 and 7.1

ivodvb:
Title speaks for itself, I saw the nightly option which currently tests for PHP 7.2.0, but I think it's better to also test 7.0 and 7.1.


rikterbeek:
Hi @ivodvb,

Thanks for the PR. 

Regards,
Rik
Adyen

PR #45, opened 2017-12-20
Check for hexadecimal digits before packing

lancergr:
**Description**
Check for hexadecimal digits before packing

**Tested scenarios**
Calculating signature with valid/invalid hmac

### DIFF
--- a/tests/Util/UtilTest.php
+++ b/tests/Util/UtilTest.php
@@ -1,0 +1,25 @@
+<?php
+
+namespace Adyen\Tests\Util;
+
+use Adyen\Util\Util;
+
+class UtilTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @expectedException        Adyen\AdyenException
+     * @expectedExceptionMessage Invalid HMAC key: INVALID
+     */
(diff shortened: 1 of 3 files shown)


tests/Util/UtilTest.php
```diff
@@ -0,0 +1,25 @@
...
+use Adyen\Util\Util;
+
+class UtilTest extends \PHPUnit_Framework_TestCase
+{
+    /**
+     * @expectedException        Adyen\AdyenException
+     * @expectedExceptionMessage Invalid HMAC key: INVALID
+     */
+    public function testSha256Invalid() {
+        Util::calculateSha256Signature("INVALID", ['key' => 'value']);
```
rikterbeek: change to array( so we can support php >=5.3.

tests/Util/UtilTest.php
```diff
@@ -0,0 +1,25 @@
...
+    }
+
+    public function testSha256() {
+        $signature = Util::calculateSha256Signature("123ABC", [
```
rikterbeek: change to array( so we can support php >=5.3.

PR #49, opened 2018-01-19
Curl init is only fire once

Klaitos:
We used your api to batch several thousands of transaction.
Using curl_init only once reduce the average time of curling.


Klaitos:
I will handle with guzzle async request in my own code.

PR #56, opened 2018-02-12
Add skipDetails.shtml endpoints to the main Client and getter and setter

rooboy35:
**Description**
I wanted to use the Adyen Client to get the endpoint for the skipDetails functionality, so have added the test/live endpoints to the Client and added a getter and setter

**Tested scenarios**
Added a basic test for both Test and Live, to getSkipDetailsUrl() to check that we get the relevant skipDetails.shtml endpoint for each.

**Fixed issue**:


rikterbeek:
Hi @rooboy35,

Thanks for the PR. I do not see a working solution here for our HPP integration. For HPP you need to do a post to the Adyen HPP environment to show the payment methods this is basically outside our Library because this contains only API calls to our server.

Regards,
Rik
Adyen

rooboy35:
The idea here was that we then wouldn't need to hardcode skipDetails endpoint URLs if already including this Adyen library. anyway, no worries

rikterbeek:
Hi rooboy35,

The problem here is to support this part of the code. The code has no useful use inside the library so it could be removed when we are optimising the code and then your implementation will fail.

Regards,
Rik
Adyen

PR #62, opened 2018-05-04
PW-421: Adding Checkout and Checkout Utility support, modified CurlCl…

Aleffio:
…ient to make it testable

**Description**
added the following endpoints:
SDK Integration:
 /setup
 /verify

API Integration:
 /paymentMethods
 /payments
 /payments/details

Checkout Utility:
 /originKeys

### DIFF
--- a/src/Adyen/Client.php
+++ b/src/Adyen/Client.php
@@ -22,8 +22,6 @@ class Client
     const TERMINAL_CLOUD_LIVE = "https://terminal-api-live.adyen.com";
     const ENDPOINT_CHECKOUT_TEST = "https://checkout-test.adyen.com";
     const ENDPOINT_CHECKOUT_LIVE = "https://checkout-live.adyen.com";
-    const ENDPOINT_CHECKOUT_UTILITY_TEST = "https://checkout-test.adyen.com/checkoutUtility";
-    const ENDPOINT_CHECKOUT_UTILITY_LIVE = "https://checkout-live.adyen.com/checkoutUtility";
 
     /**
      * @var Adyen_Config $config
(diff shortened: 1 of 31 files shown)


src/Adyen/Service/Checkout.php
```diff
@@ -0,0 +1,58 @@
...
+    public function __construct(\Adyen\Client $client)
+    {
+        parent::__construct($client);
+
+        $this->_setup = new \Adyen\Service\ResourceModel\Checkout\Setup($this);
+        $this->_verify = new \Adyen\Service\ResourceModel\Checkout\Verify($this);
+        $this->_paymentMethods = new \Adyen\Service\ResourceModel\Checkout\PaymentMethods($this);
+        $this->_payments = new \Adyen\Service\ResourceModel\Checkout\Payments($this);
+        $this->_paymentsDetails = new \Adyen\Service\ResourceModel\Checkout\PaymentsDetails($this);
+        $this->_supportsXAPIKey = true;
```
lancergr: now that the pal is accepting apiKey, we should better change it to "requiresApiKey".
Preferably bring another \Adyen\Service subclass (eg. ApiKeyAuthenticatedService) that has it set by default and let Checkout extending the new service class

src/Adyen/Client.php
```diff
@@ -8,16 +8,22 @@
...
+    const API_VERSION = "v30";
+    const API_RECURRING_VERSION = "v25";
+    const API_CHECKOUT_VERSION = "v32";
+    const API_CHECKOUT_UTILITY_VERSION = "v1";
+    const TERMINAL_CLOUD_TEST = "https://terminal-api-test.adyen.com";
+    const TERMINAL_CLOUD_LIVE = "https://terminal-api-live.adyen.com";
+    const ENDPOINT_CHECKOUT_TEST = "https://checkout-test.adyen.com";
+    const ENDPOINT_CHECKOUT_LIVE = "https://checkout-live.adyen.com";
+    const ENDPOINT_CHECKOUT_UTILITY_TEST = "https://checkout-test.adyen.com/checkoutUtility";
+    const ENDPOINT_CHECKOUT_UTILITY_LIVE = "https://checkout-live.adyen.com/checkoutUtility";
```
rikterbeek: remove those two lines.

tests/MockTest/CheckoutTest.php
```diff
@@ -0,0 +1,281 @@
...
+        // initialize service
+        $service = new \Adyen\Service\Checkout($client);
+
+        $params = array('merchantAccount' => "YourMerchantAccount");
+        try {
+            $result = $service->paymentMethods($params);
+        } catch (\Exception $e) {
+        }
+
+        $this->assertEquals('Adyen\AdyenException', get_class($e));
```
lancergr: better put these exception assertions within catch and add a fail() call after line 58

PR #74, opened 2018-10-02
PW-645 use dynamic identifier for live endpoint

rikterbeek:
PW-645 use dynamic identifier for live if provided, not provided throw error for checkout service as this is mandatory

**Description**

**Tested scenarios**

Added checkout live test for this.

**Fixed issue**:

### DIFF
--- a/src/Adyen/Client.php
+++ b/src/Adyen/Client.php
@@ -12,16 +12,18 @@ class Client
     const USER_AGENT_SUFFIX = "adyen-php-api-library/";
     const ENDPOINT_TEST = "https://pal-test.adyen.com";
     const ENDPOINT_LIVE = "https://pal-live.adyen.com";
+    const ENDPOINT_LIVE_SUFFIX = "-pal-live.adyenpayments.com";
     const ENDPOINT_TEST_DIRECTORY_LOOKUP = "https://test.adyen.com/hpp/directory/v2.shtml";
     const ENDPOINT_LIVE_DIRECTORY_LOOKUP = "https://live.adyen.com/hpp/directory/v2.shtml";
     const API_VERSION = "v30";
     const API_RECURRING_VERSION = "v25";
     const API_CHECKOUT_VERSION = "v32";
     const API_CHECKOUT_UTILITY_VERSION = "v1";
-    const TERMINAL_CLOUD_TEST = "https://terminal-api-test.adyen.com";
-    const TERMINAL_CLOUD_LIVE = "https://terminal-api-live.adyen.com";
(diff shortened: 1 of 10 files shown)


src/Adyen/Client.php
```diff
@@ -89,24 +91,32 @@ public function setXApiKey($xApiKey)
...
-            $this->_config->set('endpointTerminalCloud', self::TERMINAL_CLOUD_LIVE);
-            $this->_config->set('endpointCheckout', self::ENDPOINT_CHECKOUT_LIVE);
+            $this->_config->set('endpointTerminalCloud', self::ENDPOINT_TERMINAL_CLOUD_LIVE);
+
+            if ($liveEndpointUrlPrefix) {
+                $this->_config->set('endpoint', self::ENDPOINT_PROTOCOL . $liveEndpointUrlPrefix . self::ENDPOINT_LIVE_SUFFIX);
+                $this->_config->set('endpointCheckout', self::ENDPOINT_PROTOCOL . $liveEndpointUrlPrefix . self::ENDPOINT_CHECKOUT_LIVE_SUFFIX);
+            } else {
+                $this->_config->set('endpoint', self::ENDPOINT_LIVE);
+                $this->_config->set('endpointCheckout', null); // not supported please specify unique identifier
```
lancergr: minor: let's not set it to null, in case they specify this value "endpointCheckout " before setEnvironment call

src/Adyen/Service/AbstractCheckoutResource.php
```diff
@@ -0,0 +1,27 @@
...
+     * @param $service
+     * @return mixed
+     * @throws \Adyen\AdyenException
+     */
+    public function getCheckoutEndpoint($service)
+    {
+        // check if endpoint is set
+        if ($service->getClient()->getConfig()->get('endpointCheckout') == null) {
+            $logger = $service->getClient()->getLogger();
+            $msg = "You forgot to specify a unique identifier when setting the environment";
```
lancergr: Maybe be more explicit like:
"Maybe you forgot to specify a unique identifier when calling Client->setEnvironment method" ?

src/Adyen/Client.php
rikterbeek: if you first provide setEnvironment TEST and override with LIVE this result in a TEST endpoint that is why i specifically set it to null. Let's verify your use case tomorrow.

src/Adyen/Service/AbstractCheckoutResource.php
rikterbeek: let's discuss this tomorrow with the team for the best error description

PR #77, opened 2018-10-10
PW-562 Add applicationInformation into PHP API library

cyattilakiss:
**Description**
Adding filterParameters in the resources where you can define if there
is any parameter in the request which you don't want to send towards the
specified API endpoint
- PHP Doc blocks added

**Tested scenarios**
Unit tests added


rikterbeek:
Hi @cyattilakiss,

I would do it the opposite way so if a new resources are added it is by default not used. So allowApplicationInfo = true on the resource where we allow it and on the AbstractResource put the default on false.

Regards,
Rik
Adyen

src/Adyen/Client.php
```diff
@@ -154,6 +155,28 @@ public function setApplicationName($applicationName)
...
 
+	/**
+	 * Set external platform name and version
+	 *
+	 * @param string $name
+	 * @param string $version
+	 */
+	public function setExternalPlatform($name, $version)
+	{
+		$this->_config->set('externalPlatform', array('name' => $name, 'version' => $version));
```
rkewlani: externalPlatform has three fields. I am seeing integrator field is missing. Is this intentional ?

PR #78, opened 2018-10-11
fix local unit test running caused by the missing date_default_timezone_set

cyattilakiss:
**Description**
Add new setting to the config/tests.ini where it can be set if you are
working on the api library standalone on local
When this setting is set true then the missing defaults that can cause
issues are being set

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -9,15 +9,33 @@ class TestCase extends \PHPUnit_Framework_TestCase
     protected $_skinCode;
     protected $_hmacSignature;
 
+	/**
+	 * Settings parsed from the config/tests.ini file
+	 *
+	 * @var array
+	 */
+    protected $settings;
+
     public function __construct()
     {
-
         $this->_merchantAccount = $this->getMerchantAccount();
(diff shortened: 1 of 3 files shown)


tests/TestCase.php
```diff
@@ -9,15 +9,36 @@ class TestCase extends \PHPUnit_Framework_TestCase
...
+		$this->setDefaultsDuringDevelopment();
     }
 
+	/**
+	 * During development of the standalone php api library this function sets the necessary defaults which would
+	 * otherwise set by the merchant
+	 */
+    private function setDefaultsDuringDevelopment()
+	{
+		if ($this->isDev()) {
```
lancergr: what's the downside on skipping the dev flag and isDev for these unit tests?

cyattilakiss: Right now nothing because the only setting we apply is the default time zone and only if the client doesn't have a default value for that, but for the future I'd say this flag will be useful
If you say that do it more agile and skip the parts that's not really necessary for us now I can remove it

lancergr: I believe is better to remove for simplicity. Unit tests are mostly considered dev env anyway.

PR #81, opened 2018-10-22
Minor changes and styling

cyattilakiss:
**Description**
In the first commit you find a few small php code changes
In the second only phpdoc and autoformatting for PSR1 and PSR2

### DIFF
--- a/src/Adyen/HttpClient/CurlClient.php
+++ b/src/Adyen/HttpClient/CurlClient.php
@@ -4,257 +4,256 @@ namespace Adyen\HttpClient;
 
 class CurlClient implements ClientInterface
 {
-    /**
-     * Json API request to Adyen
-     *
-     * @param \Adyen\Service $service
-     * @param $requestUrl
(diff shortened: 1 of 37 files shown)


src/Adyen/HttpClient/CurlClient.php
```diff
@@ -2,281 +2,282 @@
...
-        $logger->info("JSON Response is: " . $result);
-
-        // result not 200 throw error
-        if ($httpStatus != 200 && $result) {
-            $this->handleResultError($result, $logger);
-        } elseif (!$result) {
-
-            list($errno, $message) = $this->curlError($ch);
-
-            curl_close($ch);
```
lancergr: curlError at L 83 will throw an exception so curl_close won't be called.
Better refactor with this sequence:
```
$response = curl_exec($curl);
$err = curl_error($curl);
curl_close($curl);
```
and then call handleResultError or handleCurlError

PR #84, opened 2018-10-31
PW-306 Prerequisite for secure fields

cyattilakiss:
**Description**
Always use api key when possible
Some factories to make Magento2 modul unit testing possible


src/Adyen/ClientFactory.php
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Adyen;
+
+class ClientFactory
+{
+	/**
+	 * @param \Adyen\Client $client
+	 * @return CheckoutUtility
```
lancergr:
```suggestion
	 * @return Client
```

PR #90, opened 2018-11-27
fix tests

rikterbeek:
**Description**

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/tests/TestCase.php
+++ b/tests/TestCase.php
@@ -18,6 +18,8 @@ class TestCase extends \PHPUnit_Framework_TestCase
 
     public function __construct()
     {
+        date_default_timezone_set('Europe/Amsterdam');
+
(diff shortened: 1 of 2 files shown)


tests/TestCase.php
```diff
@@ -18,6 +18,8 @@ class TestCase extends \PHPUnit_Framework_TestCase
 
     public function __construct()
     {
+        date_default_timezone_set('Europe/Amsterdam');
```
cyattilakiss: $this->setDefaultsDuringDevelopment(); at the bottom of this constructor does the same, doesnt't it? If it's not working then maybe that should be called first

rikterbeek: yes will remove this.

rikterbeek:
```suggestion
```

PR #108, opened 2019-06-18
PW-911 Create ConnectedTerminals request into PHP Library

cyattilakiss:
**Tested scenarios**
Successful API call with response code 200


src/Adyen/Service/PosPayment.php
```diff
@@ -19,6 +19,11 @@ class PosPayment extends \Adyen\ApiKeyAuthenticatedService
 	 */
 	protected $txType;
 
+    /**
+     * @var ResourceModel\Payment\TerminalCloudAPI
```
Aleffio:
```suggestion
     * @var ResourceModel\Payment\ConnectedTerminals
```

VictorMetzler:
Hi guys, I'm do not know if this is the right spot do write but I've a question. Where do I find in this API the methods do talk with POS Terminals ? the only thing that I've found was this method ( getConnectedTerminals ). Ho do I, for example, send a payment for a terminal ?

PR #112, opened 2019-07-17
PW-1442 Update adyen-php-api-library to checkout v49

cyattilakiss:
**Description**
Bump version to 3.0.0
Update checkout api version from 41 to 49

### DIFF
--- a/src/Adyen/Client.php
+++ b/src/Adyen/Client.php
@@ -8,7 +8,7 @@ use Monolog\Handler\StreamHandler;
 
 class Client
 {
-    const LIB_VERSION = "2.1.0";
+    const LIB_VERSION = "3.0.0";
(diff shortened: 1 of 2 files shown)


cyattilakiss:
> Before releasing make sure our Magento extension will not be using this version as it will change payment method behaviour. As well we need to make sure in the release notes we mention the impact of changes for certain payment methods moving from v41 to v49 please talk to me before releasing.

The M2 plugin is already updated in the next release: https://github.com/Adyen/adyen-magento2/pull/478/files#diff-b5d0ee8c97c7abd7e3fa29b9a27d1780R17

I also got the changelog from checkout for v49 and commented under the ticket: PW-1442

Let's discuss it on Friday :)

PR #113, opened 2019-07-18
PW-1520: added formatAmount function and test

Aleffio:
added formatAmount function and test
fix ExceptionTest


tests/Util/UtilTest.php
```diff
@@ -22,4 +22,11 @@ public function testSha256() {
         ));
         $this->assertEquals("YtbpYcrdbvk0RSVwTwENMzomS0LYtiItMwXhI5tohXs=", $signature);
     }
+
+    public function testFormatAmount() {
+        $amount = 15.02;
+        $currency = "EUR";
+        $formattedAmount = Util::formatAmount($amount, $currency);
+        $this->assertEquals(1502, $formattedAmount);
+    }
```
rikterbeek: shall we add 3 test cases for the 3 different options ?

Aleffio: done

PR #115, opened 2019-07-24
Adding ability to override the API Checkout Version

rooboy35:
**Description**
To integrate the relatively new Adyen Dropin/Components functionality, a minimum API version of 'v49' is required, however this is currently a constant and not configurable in this SDK.
The fix here is to keep the default constant as-is, but to be able to call `$client->setApiCheckoutVersion("v49")` dynamically.

**Tested scenarios**
New test `CheckoutTest->testApiCheckoutVersionOverride()` which asserts:
- the current API version being used matches `Client::API_CHECKOUT_VERSION` by default
- if the new `setApiCheckoutVersion()` function is called, retrieving the active version from the client matches the new value that was specified

Also fixed some broken tests in `tests/TestCase.php`

**Fixed issue**:  
No issue was reported before this change


cyattilakiss:
Hi @rooboy35 ,

Thank you for your PR, we are going to review and discuss your approach in the upcoming sprint.

Best,
Attila
Adyen

mr-feek:
I was also looking for a way to override this version 👍

rikterbeek:
Hi @rooboy35,

Thanks for the PR. As we want to have control what API version is used by what merchants as it is not always backwards compatible with all the payments methods we will not allow this PR. You are fully able to do a PR for just updating the version to v49 so we can review and do additional changes if needed. We have done this now already and we make a new release this week so you can update to the latest version.

Regards,
Rik
Adyen

mr-feek:
What are the implications of this API version differing from the version listed on the "API URLs and response" page within the account settings?

PR #117, opened 2019-08-27
PW-1532: Add HMAC validation on all libraries Add method to validate …

Aleffio:
…hmac

**Description**

**Tested scenarios**

**Fixed issue**:

### DIFF
--- a/src/Adyen/Util/Util.php
+++ b/src/Adyen/Util/Util.php
@@ -77,4 +77,17 @@ class Util
 
         return (int)number_format($amount, $format, '', '');
     }
+
+    public static function isValidHmac($params, $key)
+    {
+        if (empty($params["additionalData"]) || empty($params["additionalData"]["hmacSignature"])) {
+            throw new \Adyen\AdyenException("You did not provide hmacSignature in additionalData");
+        }
+        $merchantSign = $params["additionalData"]["hmacSignature"];
+        unset($params["additionalData"]);
+        $expectedSign = Util::calculateSha256Signature($key, $params);
+
+        return $expectedSign == $merchantSign;
+    }
+
 }
--- a/src/Adyen/Util/Util.php
+++ b/src/Adyen/Util/Util.php
@@ -103,18 +103,24 @@ class Util
 
     public static function getNotificationDataToSign($params)
     {
-        $value = $params['amount']['value'];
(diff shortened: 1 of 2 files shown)


msilvagarcia:
Not sure why Travis CI is giving an error. My main hypothesis is that Travis has a more strict configuration than our local servers.

Talked with @Aleffio and found a problem with the algorithm, where we are assuming the value is going to be converted to a proper string but it comes as a plain `Array` string.

src/Adyen/Util/Util.php
```diff
@@ -78,14 +78,69 @@ public static function formatAmount($amount, $currency)
...
+            $merchantReference,
+            $value,
+            $currency,
+            $eventCode,
+            $success
+        );
+
+        // The character escape function
+        $escapeval = function ($val) {
+            return str_replace(':', '\\:', str_replace('\\', '\\\\', $val));
```
msilvagarcia: We can replace this with `preg_replace('/(:|\\)/g', '\\$1', $val)` or something like this. This way we remove the nested function calls, but I understand that it adds regular expressions (and a not very readable one).

Another option would be to call them in two separate lines instead.

src/Adyen/Util/Util.php
```diff
@@ -78,14 +78,69 @@ public static function formatAmount($amount, $currency)
...
+    public static function getNotificationDataToSign($params)
+    {
+        $pspReference = (!empty($params['pspReference'])) ? $params['pspReference'] : "";
+        $originalReference = (!empty($params['originalReference'])) ? $params['originalReference'] : "";
+        $merchantAccountCode = (!empty($params['merchantAccountCode'])) ? $params['merchantAccountCode'] : "";
+        $merchantReference = (!empty($params['merchantReference'])) ? $params['merchantReference'] : "";
+        $value = (!empty($params['amount']['value'])) ? $params['amount']['value'] : "";
+        $currency = (!empty($params['amount']['currency'])) ? $params['amount']['currency'] : "";
+        $eventCode = (!empty($params['eventCode'])) ? $params['eventCode'] : "";
+        $success = (!empty($params['success'])) ? $params['success'] : "";
```
msilvagarcia: This sequence of lines don't look very scalable. How often do we add new parameters to this string?

PR #119, opened 2019-09-01
Allow the library to be used with Monolog 2

krisbuist:
**Description**
Monolog has [released v2.0.0](https://github.com/Seldaek/monolog/releases/tag/2.0.0), but the Adyen PHP library blocks upgrading to this version.

Because Monolog 2.0.0 requires at least PHP 7.2, the version constraint for this library can't just be upped to v2, because that would drop support for PHP5.3

With the `>=1.16` users have the possibility of choosing the version of Monolog which suits them

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
   "license": "Apache-2.0",
   "require": {
     "php": ">=5.3",
-    "monolog/monolog": "~1.16"
+    "monolog/monolog": ">=1.16"
   },
   "require-dev": {
     "phpunit/phpunit": "~4",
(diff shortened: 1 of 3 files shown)


rikterbeek:
Hi @krisbuist,

Thanks for the PR! Your changes break continuous test cases for PHP 7.2 and PHP 7.3 can you fix that ?

Regards,
Rik
Adyen

krisbuist:
Hi @rikterbeek. Ah you're right! Somehow I missed that.
I hope the last commit fixed the issue without changing the intention of the test. Unfortunately, most tests are skipped on my local machine.

Issue #121, opened 2019-09-03
Idempotency-Key support


zggonzalez:
**PHP version**: 7.0
**Library version**: 3.0.0
**Description**

- please add support for the header Idempotency-Key:<key> in the library

msilvagarcia:
Hey @zggonzalez ! Thanks for the suggestion. We'll plan adding this feature in the near future. :)

Cheers!
Marcos
Adyen

zggonzalez:
Hi guys, 

We're having few issues with duplicated payment authorizations. Do you have a timeline for this? Just to understand if I need to implement the idempotency in a different way or I can wait for your API release?

Thx in advance.

cyattilakiss:
Hi @zggonzalez 

In the #124 the idempotency key support is going to be merged soon.

Best,
Attila
Adyen

cyattilakiss:
The fix is merged, I'm going to close this issue.

Best,
Attila

PR #124, opened 2019-10-01
[PW-1673] add support for idempotency

rikterbeek:
**Description**

**Tested scenarios**

**Fixed issue**:  
#121

### DIFF
--- a/src/Adyen/HttpClient/CurlClient.php
+++ b/src/Adyen/HttpClient/CurlClient.php
@@ -16,7 +16,7 @@ class CurlClient implements ClientInterface
 	 * @return mixed
 	 * @throws AdyenException
 	 */
-	public function requestJson(\Adyen\Service $service, $requestUrl, $params)
+	public function requestJson(\Adyen\Service $service, $requestUrl, $params, $requestOptions = null)
 	{
 		$client = $service->getClient();
 		$config = $client->getConfig();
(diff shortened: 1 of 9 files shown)


src/Adyen/HttpClient/CurlClient.php
```diff
@@ -51,6 +51,11 @@ public function requestJson(\Adyen\Service $service, $requestUrl, $params)
 			'User-Agent: ' . $userAgent
 		);
 
+		// if idempotency key is provided as option include into request
+        if ($requestOptions != null && !empty($requestOptions['idempotencyKey'])) {
```
cyattilakiss:
```suggestion
        if (!empty($requestOptions['idempotencyKey'])) {
```

Checking if null is not necessary - empty will take care of that

src/Adyen/Util/Uuid.php
```diff
@@ -0,0 +1,48 @@
...
+{
+
+    /**
+     * Generate a UUID V4
+     *
+     * @return string
+     */
+    public static function generateV4()
+    {
+        $random = openssl_random_pseudo_bytes(32, $isSourceStrong);
```
cyattilakiss: Isn't the $isSourceStrong missing?

rikterbeek: no see docs: If passed into the function, this will hold a boolean value that determines if the algorithm used was "cryptographically strong", e.g., safe for usage with GPG, passwords, etc. TRUE if it did, otherwise FALSE
https://www.php.net/manual/en/function.openssl-random-pseudo-bytes.php

Issue #127, opened 2019-10-09
202 HTTP status code treated as errors


fredericbarthelet:
Hi,

Your current library indicates that any other response than HTT status code 200 result in an error.

https://github.com/Adyen/adyen-php-api-library/blob/c0d7a1db0634d95e6fdeb57e369d7161648bde9d/src/Adyen/HttpClient/CurlClient.php#L96

However your documentation states that interactions with MarketPay Account endpoint may results in [200 or 202](https://docs.adyen.com/marketpay/onboarding-and-verification) 

![image](https://user-images.githubusercontent.com/23530626/66462119-b1b74b00-ea7a-11e9-93bb-48837a83c05c.png)

If you want I'll produce the PR to handle every 2xx codes as OK codes.

msilvagarcia:
Hey Frederic,

Thanks for raising the issue. If you could produce the pull request, that would be great.

Cheers!
Marcos
Adyen

fredericbarthelet:
Hey @msilvagarcia, made a PR for this issue, could you review it ? Thanks

cyattilakiss:
Hi @fredericbarthelet ,

I will close this ticket since the PR has been merged.
Please feel free to reopen if the issue persist.

Best,
Attila
Adyen

PR #130, opened 2019-10-18
Add 2xx and 3xx response code as valid status codes

fredericbarthelet:
**Description**
Accept 2xx and 3xx status code response. This aims to handle 202 status code not as error.

**Tested scenarios**
Account creation

**Fixed issue**:
#127

### DIFF
--- a/src/Adyen/HttpClient/CurlClient.php
+++ b/src/Adyen/HttpClient/CurlClient.php
@@ -202,7 +202,9 @@ class CurlClient implements ClientInterface
 
 		curl_close($ch);
 
-		if ($httpStatus >= 400 && $result) {
+		$resultOKHttpStatusCodes = array(200, 201, 202, 204);
+		
+		if (!in_array($httpStatus, $resultOKHttpStatusCodes) && $result) {
 			$this->handleResultError($result, $logger);
 		} elseif (!$result) {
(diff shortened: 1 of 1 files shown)


msilvagarcia:
Hey @fredericbarthelet !

Can you explain the reason of accepting 3xx status codes as well? Also, what happens when a status code 1xx is received?

src/Adyen/HttpClient/CurlClient.php
```diff
@@ -202,7 +202,7 @@ public function requestPost(\Adyen\Service $service, $requestUrl, $params)
 
 		curl_close($ch);
 
-		if ($httpStatus != 200 && $result) {
+		if ($httpStatus >= 400 && $result) {
```
cyattilakiss:
```suggestion
		$resultOKHttpStatusCodes = array(200, 201, 202, 203, 204);
		
		if (!in_array($httpStatus, $resultOKHttpStatusCodes) && $result) {
```

I would keep it "simple" and consider OK only the known response codes that can be returned from the API.

fredericbarthelet:
@cyattilakiss PR updated as per your suggestion :)

cyattilakiss:
Hi @fredericbarthelet ,

Thanks it will be merged :)

Best,
Attila

PR #134, opened 2019-11-01
Add composer.lock

msilvagarcia:
This standardizes the versions of libraries used.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,9 +4,6 @@ tests/config
 # Ignore the /vendor/ directory
 /vendor/
 
-# composer.lock file should be ignored
-composer.lock
-
 # ignore Mac OS X file
 .DS_Store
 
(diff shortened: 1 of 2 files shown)


rikterbeek:
@msilvagarcia what is the benefit of this ?

msilvagarcia:
@rikterbeek From the [composer documentation](https://getcomposer.org/doc/01-basic-usage.md#commit-your-composer-lock-file-to-version-control):

> Committing this file to [Version Control] is important because it will cause anyone who sets up the project to use the exact same versions of the dependencies that you are using. Your [Continuous Integration] server, production machines, other developers in your team, everything and everyone runs on the same dependencies, which mitigates the potential for bugs affecting only some parts of the deployments. Even if you develop alone, in six months when reinstalling the project you can feel confident the dependencies installed are still working even if your dependencies released many new versions since then.

I think it's a nice safe-guard to have in place. Might prevent some headaches.

PR #147, opened 2020-01-09
Fix bad phpdoc typehint for $liveEndpointUrlPrefix argument

ikvasnica:
**Description**
Argument `$liveEndpointUrlPrefix` has a wrong PHPDoc type hint `null`, whereas `string|null` is expected in the code.

**Fixed issue**:  Static analysis tools, such as PHPStan or Phan, rely on PHPDoc and fail due to wrong type hint.

### DIFF
--- a/src/Adyen/Client.php
+++ b/src/Adyen/Client.php
@@ -124,7 +124,7 @@ class Client
      * For live please specify the unique identifier.
      *
      * @param string $environment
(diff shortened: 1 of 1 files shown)


cyattilakiss:
Hi @ikvasnica ,

Thank you for opening this PR. We are going to merge and release it in our next release.

Best,
Attila
Adyen

PR #160, opened 2020-01-28
update notification endpoint version

cangelis:
**Description**

Update Notification endpoint version.

`v1` version doesn't seem to be saving the `eventConfigs`

### DIFF
--- a/src/Adyen/Client.php
+++ b/src/Adyen/Client.php
@@ -22,7 +22,7 @@ class Client
     const API_RECURRING_VERSION = "v25";
     const API_CHECKOUT_VERSION = "v51";
     const API_CHECKOUT_UTILITY_VERSION = "v1";
-    const API_NOTIFICATION_VERSION = "v1";
+    const API_NOTIFICATION_VERSION = "v5";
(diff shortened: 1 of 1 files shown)


rikterbeek:
Hi @cangelis,

Thanks for the PR! We will look if this will result in a breaking change or not so we can create the right release number for this. When we know this we will merge and provide a new release.

Regards,
Rik
Adyen

cangelis:
Hi @rikterbeek 

This is a major blocker for us to go live soon and the existing `v1` endpoint doesn't meet the official api documentation.

I thought this library is the official php library of adyen and we rely on the library from day 1 because of this reason (which was a huge mistake) but you can't be aware that if an endpoint version change can cause a BC break instantly. Also many features are missing and we also try to contribute to the project at the same time.

I'd be appreciated if you can look into this asap since we are about to go live soon.

rikterbeek:
Hi @cangelis,

Understand, we will merge this PR and create a new major release for this library. We will update all APIs to latest version.

Regards,
Rik
Adyen

PR #164, opened 2020-02-04
[PW-2059] Mask encrypted/sensitive data in the logs

cyattilakiss:
**Description**
Mask the encrypted card details and other sensitive data from the logs of the payments requests and responses if the environment is live

### DIFF
--- a/src/Adyen/Config.php
+++ b/src/Adyen/Config.php
@@ -165,4 +165,12 @@ class Config implements ConfigInterface
     {
         return isset($this->data['externalPlatform']) ? $this->data['externalPlatform'] : null;
     }
+
+    /**
+     * @return mixed|null
+     */
+    public function getEnvironment()
+    {
+        return isset($this->data['environment']) ? $this->data['environment'] : null;
+    }
 }
--- a/src/Adyen/Config.php
+++ b/src/Adyen/Config.php
@@ -167,7 +167,7 @@ class Config implements ConfigInterface
     }
 
     /**
-     * @return mixed|null
+     * @return string|null
      */
     public function getEnvironment()
     {
--- a/src/Adyen/HttpClient/CurlClient.php
+++ b/src/Adyen/HttpClient/CurlClient.php
@@ -6,6 +6,40 @@ use Adyen\AdyenException;
 
 class CurlClient implements ClientInterface
 {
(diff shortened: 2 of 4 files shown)


src/Adyen/Config.php
```diff
@@ -165,4 +165,12 @@ public function getExternalPlatform()
     {
         return isset($this->data['externalPlatform']) ? $this->data['externalPlatform'] : null;
     }
+
+    /**
+     * @return mixed|null
```
msilvagarcia:
```suggestion
     * @return string|null
```

src/Adyen/HttpClient/CurlClient.php
```diff
@@ -292,23 +289,114 @@ protected function handleResultError($result, $logger)
...
+                    'expiryYear',
+                    'cvc',
+                    'encryptedCardNumber',
+                    'encryptedExpiryMonth',
+                    'encryptedExpiryYear',
+                    'encryptedSecurityCode',
+                    'applepay.token',
+                    'paywithgoogle.token'
+                )
+            );
```
msilvagarcia: Could this be a constant or property?

src/Adyen/HttpClient/CurlClient.php
```diff
@@ -292,23 +289,114 @@ protected function handleResultError($result, $logger)
...
+        // Filter sensitive data from logs when live
+        if (\Adyen\Environment::LIVE == $environment) {
+            // List of parameters that needs to be masked with the same array structure as it appears in
+            // the response array
+            $paramsToMask = array(
+                'paymentData',
+                'action' => array(
+                    'paymentData'
+                )
+            );
```
msilvagarcia: Could this be another constant or property?

src/Adyen/HttpClient/CurlClient.php
```diff
@@ -385,13 +385,23 @@ private function maskParametersRecursive($paramsToMaskList, $params)
...
+     *
      * @param $parameter
      * @return string
      */
     private function maskParameter($parameter)
     {
-        if (!empty($parameter)) {
+        if (empty($parameter)) {
+            return $parameter;
+        }
```
msilvagarcia: I like the early return :)

PR #144, opened 2019-12-24
add /getCostEstimate Resource model

cangelis:
**Description**
Add missing `/getCostEstimate` endpoint resource model

PR #171, opened 2020-02-13
Release 6.0.0

msilvagarcia:
This releases version 6.0.0 of the Adyen PHP API Library

Remarkable features include:
- Add /getCostEstimate Resource model (#144)
- PW-1996 Remove escaping from notification HMAC (#165)
- Update to latest API version (#162)
- PW-2059 Mask encrypted/sensitive data in the logs (#164)
- Update notification endpoint version (#160)

Because of the update to the latest API version (#162), this is a major release. There are changes to the MarketPay configuration API responses that users need to address after updating.


src/Adyen/HttpClient/CurlClient.php
```diff
@@ -1,57 +1,92 @@
...
+    // the request array
+
+    private static $requestParamsToMask = array(
+        'paymentData',
+        'card' => array(
+            'number',
+            'cvc'
+        ),
+        'additionalData' => array(
+            'card.encrypted.json',
```
Aleffio:
```suggestion
            'card.encrypted.json'
```

src/Adyen/Service/ResourceModel/BinLookup/GetCostEstimate.php
```diff
@@ -0,0 +1,17 @@
+<?php
+
+namespace Adyen\Service\ResourceModel\BinLookup;
+
+class GetCostEstimate extends \Adyen\Service\AbstractResource
+{
```
Aleffio:
```suggestion
{
    /**
     * @var string
     */
    protected $endpoint;
    
    
    /**
     * Include applicationInfo key in the request parameters
     *
     * @var bool
     */
    protected $allowApplicationInfo = false;
    
```

Let's do like the other services?

src/Adyen/Service/ResourceModel/BinLookup/GetCostEstimate.php
```diff
@@ -0,0 +1,17 @@
...
+    /**
+     * GetCostEstimate constructor.
+     *
+     * @param \Adyen\Service $service
+     */
+    public function __construct($service)
+    {
+        $endpoint = $service->getClient()->getConfig()->get('endpoint') . '/pal/servlet/BinLookup/' . $service->getClient()->getApiBinLookupVersion() . '/getCostEstimate';
+        parent::__construct($service, $endpoint, false);
```
Aleffio:
```suggestion
        parent::__construct($service, $this->endpoint, $this->allowApplicationInfo);
```

src/Adyen/Service/ResourceModel/BinLookup/GetCostEstimate.php
```diff
@@ -0,0 +1,17 @@
...
+class GetCostEstimate extends \Adyen\Service\AbstractResource
+{
+    /**
+     * GetCostEstimate constructor.
+     *
+     * @param \Adyen\Service $service
+     */
+    public function __construct($service)
+    {
+        $endpoint = $service->getClient()->getConfig()->get('endpoint') . '/pal/servlet/BinLookup/' . $service->getClient()->getApiBinLookupVersion() . '/getCostEstimate';
```
Aleffio:
```suggestion
        $this->endpoint = $service->getClient()->getConfig()->get('endpoint') . '/pal/servlet/BinLookup/' . $service->getClient()->getApiBinLookupVersion() . '/getCostEstimate';
```

msilvagarcia:
@Aleffio while I believe your suggestions are great, I don't think this Pull Request is the best place to apply them. This would mean we're pushing commits to the develop branch, which should only be done with Pull Requests.

Do you think I should open a new pull request for these changes?

Aleffio:
@msilvagarcia I agree this is not the best place, though this review is like a "second chance" of seeing something before it goes to the master branch. I don't believe these suggestions are so complex to need a separate pull request, but feel free to open one if you think it is needed

PR #182, opened 2020-04-15
Update composer.lock

msilvagarcia:
Renovate bot missed this on pull request #180. Manually updated it for now. 

Rewrote several tests to make it work.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -4,25 +4,25 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "bf68c7d2c60f989124cefbf1b5c38762",
+    "content-hash": "f25379c30bb450a8b004f29f764a0099",
     "packages": [
         {
             "name": "monolog/monolog",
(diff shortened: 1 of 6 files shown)


composer.lock
```diff
@@ -4,59 +4,62 @@
...
-                "url": "https://api.github.com/repos/Seldaek/monolog/zipball/fa82921994db851a8becaf3787a9e73c5976b6f1",
-                "reference": "fa82921994db851a8becaf3787a9e73c5976b6f1",
+                "url": "https://api.github.com/repos/Seldaek/monolog/zipball/c861fcba2ca29404dc9e617eedd9eff4616986b8",
+                "reference": "c861fcba2ca29404dc9e617eedd9eff4616986b8",
                 "shasum": ""
             },
             "require": {
-                "php": ">=5.3.0",
-                "psr/log": "~1.0"
+                "php": "^7.2",
```
cyattilakiss, line 24: I think we should keep >= 5.6 for this so we need to lower the monolog/monolog version :(
As far as I see this is not a dev dependency

msilvagarcia: nice catch!

PR #192, opened 2020-05-08
[PW-2248]: Outsourcing of methods to library

acampos1916:
**Description**
We're moving some classes and methods from the PrestaShop repo to the library so they can be used on all plugins.

**Tested scenarios**
None

**Fixed issue**:  Part of PW-2248


composer.json
```diff
@@ -8,8 +8,11 @@
   "homepage": "https://github.com/Adyen/adyen-php-api-library",
   "license": "Apache-2.0",
   "require": {
-    "php": ">=5.3",
-    "monolog/monolog": ">=1.16"
+    "php": ">=5.4",
```
acampos1916, line 11: in order to support the hex2bin() function

cyattilakiss: This would be a breaking change so for now let's not move the function which requires this function :(

src/Adyen/Service/Builder/Browser.php
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * Copyright (c) 2020 Adyen N.V.
+ * This file is open source and available under the MIT license.
+ * See the LICENSE file for more info.
+ *
+ */
+
+namespace Adyen\Service\Builder;
+
+class Browser
```
cyattilakiss, line 26: let's remove this builder as well because I'm not sure if we need a builder for it or we can just use the component's state data after validating it - and then we don't need to build anything

src/Adyen/Service/Builder/OpenInvoice.php
```diff
@@ -0,0 +1,85 @@
...
+ *
+ * Copyright (c) 2020 Adyen N.V.
+ * This file is open source and available under the MIT license.
+ * See the LICENSE file for more info.
+ *
+ */
+
+namespace Adyen\Service\Builder;
+
+use Tools;
```
cyattilakiss, line 26: remove

src/Adyen/Service/Builder/OpenInvoice.php
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * For Klarna And AfterPay use Vat category High others use none
+     *
+     * @param $paymentMethod
+     * @return string 'High'/'None'
+     */
+    public function getVatCategory($paymentMethod)
+    {
+        if ($paymentMethod == "klarna" ||
+            Tools::strlen($paymentMethod) >= 9 &&
```
cyattilakiss, line 77: be careful with Tools -> use php function instead

src/Adyen/Service/Builder/Payment.php
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+     * @param $paymentMethodType
+     * @param $storedPaymentMethodId
+     * @param $origin
+     * @param string $encryptedSecurityCode
+     * @param array $request
+     * @return array
+     */
+    public function buildStoredPaymentData(
```
cyattilakiss, line 127: remove this function because of the component's state data

src/Adyen/Service/Builder/Payment.php
```diff
@@ -0,0 +1,163 @@
...
+     * @param $encryptedExpiryYear
+     * @param $holderName
+     * @param $origin
+     * @param string $encryptedSecurityCode
+     * @param string $paymentMethodType
+     * @param bool $storePaymentMethod
+     * @param array $request
+     * @return array
+     */
+    public function buildCardData(
```
cyattilakiss, line 40: remove this function because of the component's state data

src/Adyen/Service/NotificationReceiver.php
```diff
@@ -0,0 +1,229 @@
...
+     * @param Configuration $configuration
+     * @param AdyenNotification $adyenNotification
+     */
+    public function __construct(
+        HmacSignature $hmacSignature,
+        $notificationHMAC,
+        $merchantAccount,
+        $notificationUsername,
+        $notificationPassword,
+        LoggerInterface $logger, //TODO outsource?
```
cyattilakiss, line 94: Loggerinterface we can outsource yes

src/Adyen/Service/NotificationReceiver.php
```diff
@@ -0,0 +1,229 @@
...
+        Configuration $configuration, //TODO outsource?
```
cyattilakiss, line 95: I would remove this and adjust the functions to receive the configuration in parameters

src/Adyen/Service/NotificationReceiver.php
```diff
@@ -0,0 +1,229 @@
...
+        AdyenNotification $adyenNotification //TODO outsource?
```
cyattilakiss, line 96: Not used in this class if I'm not mistaken, can be removed!

cyattilakiss:
@msilvagarcia I see we have codacy configured here, shouldn't we have sonarcloud instead?

composer.json
```diff
@@ -9,7 +9,10 @@
   "license": "Apache-2.0",
   "require": {
     "php": ">=5.6",
-    "monolog/monolog": "^1.16"
+    "monolog/monolog": ">=1.16",
```
msilvagarcia, line 12:
```suggestion
    "monolog/monolog": "^1",
```

msilvagarcia: If you approve this, remember to run `composer update` to get a new `composer.lock` file

cyattilakiss: do we need to restrict the usage of monolog v2? I think we can let them use version 1 and 2 as well.

AlexandrosMor: @msilvagarcia ?

AlexandrosMor:
@msilvagarcia  Some platforms do not have the crypt functionality built in. For magento there is a native to use but for Shopware does not have one,

msilvagarcia:
@AlexandrosMor I don't think it should be the library's responsibility to provide filler functionality to platforms. The library itself does not require this kind of encryption, it's just necessary to store values in the platforms' databases for now.

If there's a clear use for that in the library itself I don't mind adding it.

PR #242, opened 2020-12-03
make full adyen error code accessible in AdyenException

patrickmyh:
**Description**
When an AdyenException was thrown (after a curl call to the Adyen API) in src/Adyen/HttpClient/CurlClient.php's handleResultError() method, the error code, returned from the API-Response, was not properly passed.

It was casted to an int and thus sometimes truncated.

**Fixed issue**:240

### DIFF
--- a/src/Adyen/AdyenException.php
+++ b/src/Adyen/AdyenException.php
@@ -22,6 +22,11 @@ class AdyenException extends Exception
     protected $pspReference;
 
     /**
+     * @var string
+     */
+    protected $adyenErrorCode;
+
+    /**
      * AdyenException constructor.
      *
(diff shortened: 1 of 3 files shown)


acampos1916 (approved):

cyattilakiss (approved):

oradwell:
this is a breaking change because Exception code is now mapped to "status" instead of "errorCode". We were using `$exception->getCode()` to handle Adyen errors, as of 8.1.0 we have to use `$exception->getAdyenErrorCode()`

With semantic versioning `$exception->getCode()` should have been kept the same until a major version release (e.g. 9.0.0)

PR #342, opened 2022-03-21
[PW-6283] Add manual capture payment methods list

AlexandrosMor:
**Description**

The payment methods that supports AUTH flow are also supporting manual capture. The list with the payment methods is added in the library so we have one method to update and to not duplicate the list across the platforms, 

**Tested scenarios**
Added a unit test for the payment methods variants

### DIFF
--- a/src/Adyen/Util/ManualCapture.php
+++ b/src/Adyen/Util/ManualCapture.php
@@ -1,0 +1,108 @@
+<?php
+/**
+ *                       ######
+ *                       ######
+ * ############    ####( ######  #####. ######  ############   ############
+ * #############  #####( ######  #####. ######  #############  #############
+ *        ######  #####( ######  #####. ######  #####  ######  #####  ######
+ * ###### ######  #####( ######  #####. ######  #####  #####   #####  ######
+ * ###### ######  #####( ######  #####. ######  #####          #####  ######
+ * #############  #############  #############  #############  #####  ######
+ *  ############   ############  #############   ############  #####  ######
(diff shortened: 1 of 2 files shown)


Morerice (approved):

src/Adyen/Util/ManualCapture.php
```diff
@@ -0,0 +1,108 @@
...
+
+    public function isManualCaptureSupported($notificationPaymentMethod): bool
+    {
+        $manualCaptureAllowed = false;
+        // For all openinvoice methods manual capture is the default
+        if ($this->openInvoice->isOpenInvoicePaymentMethod($notificationPaymentMethod)) {
+            return true;
+        }
+
+        switch ($notificationPaymentMethod) {
```
peterojo, line 46: I think it would be better to define a static array of constants for these methods

src/Adyen/Util/ManualCapture.php
```diff
@@ -0,0 +1,108 @@
...
+            'boleto',
+            'clearpay',
+            'afterpay',
+            'ratepay',
+            'zip',
+        );
+
+        //check the payment methods with variants
+        foreach ($paymentMethodsWithVariants as $paymentMethod) {
+            if (str_contains($notificationPaymentMethod, $paymentMethod)) {
```
peterojo, line 102: is there a reason to *not* use `in_array()`?

peterojo (approved):

Morerice (approved):

PR #359, opened 2022-05-02
[PW-6598] Add query parameters to get merchant call

AlexandrosMor:
**Description**

Add to /merchants endpoint call query params using an associative array
- Query parameters that could be used 
1. `pageNumber` 
The number of the page to fetch.
2. `pageSize` 
The number of items to have on a page, maximum 100. The default is 10 items on a page

**Tested scenarios**
End to end test added for pageSize

### DIFF
--- a/src/Adyen/Service/ResourceModel/Management/MerchantAccount.php
+++ b/src/Adyen/Service/ResourceModel/Management/MerchantAccount.php
@@ -13,12 +13,16 @@ class MerchantAccount extends \Adyen\Service\AbstractResource
     protected $allowApplicationInfo = false;
 
     /**
+     * @param array|null $queryParams
      * @return mixed
      * @throws \Adyen\AdyenException
      */
-    public function list()
+    public function list(array $queryParams = null)
     {
         $url = $this->managementEndpoint . "/merchants";
+        if (!empty($queryParams)) {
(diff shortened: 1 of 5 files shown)


src/Adyen/Service/ResourceModel/Management/MerchantAccount.php
```diff
@@ -13,12 +13,16 @@ class MerchantAccount extends \Adyen\Service\AbstractResource
...
      * @throws \Adyen\AdyenException
      */
-    public function list()
+    public function list(array $queryParams = null)
     {
         $url = $this->managementEndpoint . "/merchants";
+        if (!empty($queryParams)) {
+            $url .= '?' . http_build_query($queryParams);
+        }
         return $this->requestHttp($url, 'get');
```
peterojo, line 26: The `requestHttp` method accepts the params as array|null and takes care of adding it to the request
```suggestion
        return $this->requestHttp($url, 'get', $queryParams);
```

peterojo: curl client does not use the params afterall, better to add it there

michaelpaul (approved):

peterojo (approved):

PR #390, opened 2022-08-17
Add CS rule for quotes

szepeviktor:
❌ CI will fail.

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -32,4 +32,5 @@
     <arg value="p"/>
 
     <rule ref="PSR2"/>
+    <rule ref="Squiz.Strings.DoubleQuoteUsage"/>
 </ruleset>


szepeviktor:
All problems can be fixed by running a single command `vendor/bin/phpcbf`

It will do an excellent job. But people may argue with it.

szepeviktor:
@jillingk I'm sad that this PR got closed.

jillingk:
Hi @szepeviktor,

Not too worry, I have done and will do some more cleaning up on this library with the code fixer! However as I was doing it manually, I decided to close this PR. Thanks a lot for contributing this way though, it is very much appreciated and we do like to encourage you to do so in the future should you want to or find any need to. 

Best, Jilling
Adyen

PR #402, opened 2022-09-12
Fix for region validation & add India to the regions list

candemiralp:
List of the valid regions updated.

### DIFF
--- a/src/Adyen/Client.php
+++ b/src/Adyen/Client.php
@@ -502,7 +502,7 @@ class Client
      */
     public function setRegion(string $region): void
     {
-        if (!array_key_exists($region, Region::VALID_REGIONS)) {
+        if (!in_array($region, Region::VALID_REGIONS)) {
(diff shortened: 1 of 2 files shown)


src/Adyen/Client.php
```diff
@@ -502,8 +502,8 @@ protected function createDefaultLogger()
      */
     public function setRegion(string $region): void
     {
-        if (!array_key_exists($region, Region::TERMINAL_API_ENDPOINTS_MAPPING)) {
-            throw new AdyenException('Trying to set invalid region!');
+        if (!array_key_exists($region, Region::VALID_REGIONS)) {
```
AlexandrosMor, line 505:
```suggestion
        if (!in_array($region, Region::VALID_REGIONS)) {
```

RokPopov (approved):

michaelpaul (approved):

AlexandrosMor (approved):

PR #409, opened 2022-09-19
[PW-7128] Add deprecation tag to Logger functions and functions using them

jillingk:
**Description**
Add deprecation tag to Logger functions and functions using them.

### DIFF
--- a/src/Adyen/Client.php
+++ b/src/Adyen/Client.php
@@ -467,7 +467,11 @@ class Client
      * Set the Logger object
      *
      * @param LoggerInterface $logger
+     *
+     * @deprecated Please do not use Logger as we will deprecate this in the
+     *             future for improvements on the library
      */
+
     public function setLogger(LoggerInterface $logger)
     {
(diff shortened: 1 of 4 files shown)


src/Adyen/Client.php
```diff
@@ -467,14 +467,21 @@ protected function createDefaultHttpClient()
      * Set the Logger object
      *
      * @param LoggerInterface $logger
+     *
+     * @deprecated Please do not use Logger as we will deprecate this in the
+     *             future for improvements on the library
      */
+
```
candemiralp, line 474:
```suggestion
```

src/Adyen/Client.php
```diff
@@ -487,6 +493,7 @@ public function getLogger()
 
     /**
      * @return Logger
+     *
```
candemiralp, line 496:
```suggestion
```

candemiralp (approved):

AlexandrosMor (approved):

PR #417, opened 2022-10-18
Feat/adyen for platforms

julian-van-drielen:
**Description**
Implemented Adyen for Platforms endpoints

### DIFF
--- a/src/Adyen/Service/AbstractPlatformsResource.php
+++ b/src/Adyen/Service/AbstractPlatformsResource.php
@@ -1,0 +1,22 @@
+<?php
+
+namespace Adyen\Service;
+
+use Adyen\PlatformsService;
+
+class AbstractPlatformsResource extends AbstractResource
+{
+    /** @var string baseUrl */
+    protected $baseUrl;
+
+    public function __construct(PlatformsService $service)
+    {
+        $this->baseUrl = $service->getBaseUrl();
+
+        parent::__construct(
+            $service,
+            $this->baseUrl,
+            true
+        );
+    }
+}
--- a/src/Adyen/Service/AbstractPlatformsResource.php
+++ b/src/Adyen/Service/AbstractPlatformsResource.php
@@ -19,4 +19,18 @@ class AbstractPlatformsResource extends AbstractResource
             true
         );
     }
+
+    /**
+     * @param string $version
+     * @param string $endpoint
+     * @param callable $callback
+     * @return mixed
+     */
+    public function simulateEndpoint(string $version, string $endpoint, callable $callback)
+    {
+        $this->endpoint = sprintf('%s/%s/%s', $this->baseUrl, $version, $endpoint);
+        $response = $callback();
+        $this->endpoint = $this->baseUrl;
+        return $response;
+    }
 }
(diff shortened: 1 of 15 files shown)


src/Adyen/Service/Transfers.php
```diff
@@ -0,0 +1,62 @@
...
+     * @throws AdyenException
+     */
+    public function __construct(Client $client)
+    {
+        parent::__construct(
+            $client,
+            $client->getConfig()->get('endpointPlatformsTransfers')
+        );
+
+        $this->transfers = new ResourceModel\Transfers\Transfers($this);
```
peterojo, line 29: Why not import this class as well with a `use` statement?

src/Adyen/Service/AbstractPlatformsResource.php
```diff
@@ -0,0 +1,36 @@
...
+
+    /**
+     * @param string $version
+     * @param string $endpoint
+     * @param callable $callback
+     * @return mixed
+     */
+    public function simulateEndpoint(string $version, string $endpoint, callable $callback)
+    {
+        $this->endpoint = sprintf('%s/%s%s', $this->baseUrl, $version, $endpoint);
```
peterojo, line 31: Could you make it so that you can have either all or none of the slashes (/) here in the  sprintf format?
i.e `sprintf('%s%s%s'...` or `sprintf('%s/%s/%s'...`

peterojo (commented):
Hi @julian-van-drielen! Thanks so much for your contribution!
Do you mind adding a code example of how you would use the lib to make a request to one of the new added endpoints?

src/Adyen/Service/AbstractPlatformsResource.php
```diff
@@ -0,0 +1,36 @@
...
+        );
+    }
+
+    /**
+     * @param string $version
+     * @param string $endpoint
+     * @param callable $callback
+     * @return mixed
+     */
+    public function simulateEndpoint(string $version, string $endpoint, callable $callback)
```
peterojo, line 29: Hi @julian-van-drielen! I think this method adds a really interesting abstraction that could be used elsewhere in the library as well. We will discuss this approach internally before making a decision about adding this. However I feel like the name `simulateEndpoint` makes it sound like a test method (threw me off at first) when it is actually use to build the endpoint for all the requests you added. We would consider renaming it, but your ideas are welcome as well.

src/Adyen/Service/AbstractResource.php
```diff
@@ -100,6 +101,88 @@ public function request($params, $requestOptions = null)
...
+     */
+    public function requestPatch($params, $requestOptions = null)
+    {
+        // convert to PHP Array if type is inputType is json
+        if ($this->service->getClient()->getConfig()->getInputType() == 'json') {
+            $params = json_decode($params, true);
+            if ($params === null && json_last_error() !== JSON_ERROR_NONE) {
+                $msg = 'The parameters in the request expect valid JSON but JSON error code found: ' .
+                    json_last_error();
+                $this->service->getClient()->getLogger()->error($msg);
```
candemiralp, line 120: Dear @julian-van-drielen, thank you for your valuable contribution to support AfP in the Adyen PHP API Library.

There is an ongoing discussion to exclude all of the logging functionality in the library itself. I will discuss it with my colleagues after that logging can safely be deleted from here.

julian-van-drielen:
Hi @peterojo @candemiralp,

Thanks for the feedback, as we were in integration phase with Adyen in the past half year I was unable to take a look at this Pull Request any more.

I will process your feedback in the next weeks when I find some time to do so.

Thanks for taking a look at it.

PR #440, opened 2023-02-14
Allow monolog v3

BackEndTea:
This lib only uses monolog to create a falllback handler to stderr. So no breaking changes for this lib, and it allows users of this lib to upgrade to v3


BackEndTea:
Closing my PR in favor of the renovate bot one

PR #446, opened 2023-03-01
[ITT-369] Templates for model generation and Makefile

jillingk:
**Description**
This PR adds mustache templates for model generation and Makefile to execute said generation. Furthermore, it adds generated checkout v70 models and some service method plus UT to test it.


src/Adyen/Service/Checkout.php
```diff
@@ -216,18 +222,33 @@ public function ordersCancel($params, $requestOptions = null)
...
         return $this->donations->request($params, $requestOptions);
     }
 
+    /**
+     * @param CreateCheckoutSessionRequest $createCheckoutSessionRequest
+     * @param array|null $requestOptions
+     * @return mixed
+     * @throws AdyenException
+     */
+    public function sessionsTypes(
```
michaelpaul, line 238: This feature is still a work in progress that should not be part of the public API yet. Move this code into unit tests.

Makefile
```diff
@@ -0,0 +1,72 @@
+openapi-generator-version:=6.3.0
+openapi-generator-url:=https://repo1.maven.org/maven2/org/openapitools/openapi-generator-cli/$(openapi-generator-version)/openapi-generator-cli-$(openapi-generator-version).jar
+openapi-generator-jar:=target/openapi-generator-cli.jar
+openapi-generator-cli:=java -jar $(openapi-generator-jar)
+
+generator:=php
+services:=Checkout
+models:=src/Adyen/Model
+output:=target/out
```
michaelpaul, line 9: Is this folder git ignored?

michaelpaul (changes requested):
Maybe we should keep these changes in a new experimental branch instead of `develop` already.

src/Adyen/Service/Checkout.php
jillingk: I thought I'd add it here so people can get an idea of how it would look like, and then remove it again in the next ticket for service automation. Although I see your point, maybe it's better leave it out of this PR for now (also the checkout models itself) and just add it in the next?

jillingk: done

PR #455, opened 2023-03-10
Itt 7/update checkout v70

wboereboom:
**Description**
- update Checkout API to v70
- Add missing endpoints for CardDetails in the payments category and all endpoints in the Recurring category
- Mark classic endpoints as deprecated


michaelpaul (approved):

src/Adyen/Service/ResourceModel/Checkout/Recurring.php
```diff
@@ -0,0 +1,33 @@
...
+     *
+     * @var bool
+     */
+    protected $allowApplicationInfo = true;
+
+
+    public function delete($recurringId, array $queryParams = [])
+    {
+        $url = $this->checkoutEndpoint . self::STORED_PAYMENT_METHODS . "/" . $recurringId;
+        return $this->requestHttp($url, 'delete', $queryParams);
```
jillingk, line 25: The ```$queryParams``` here won't actually work as the requestHttp in the curlclient only reads query params for a get request. Perhaps you can add a statement in the curlclient/requestHttp to also take query parameters if the request method is a delete.

wboereboom: Good catch. Would there be any reason to keep checking for the http method?

jillingk: Yeah we do have to check it, because the params variable also get's used for body params in POST/PATCH. This is exactly why I completely copied and modified this whole method in my automation PR's.

jillingk (approved):

AlexandrosMor (approved):

PR #457, opened 2023-03-20
[ITT-398] Final Changes to prepare for Model/Service Automation

jillingk:
**Description**
This PR adds the final changes to the client/service/curl classes to prepare for the automation. It also changed the service templates further so this requires no manual work anymore (base url creation). Finally it fixed some last bugs that I ran into.


templates/model_enum.mustache
```diff
@@ -1,6 +1,6 @@
 class {{classname}}
 {
-    /**
+    /** thisrighthereasdf
```
michaelpaul, line 3: Debugging statement. :)

src/Adyen/Service.php
```diff
@@ -43,11 +43,51 @@ public function getClient()
...
+     */
+    protected function createBaseUrl(string $url): string
+    {
+        $config = $this->getClient()->getConfig();
+        if ($this->getClient()->getConfig()->get('enableLive')) {
+            // replace test in string with live
+            $url = str_replace('test', 'live', $url);
+
+            // add live url prefix if needed
+            if (str_contains($url, "checkout") || str_contains($url, "pal")) {
```
michaelpaul, line 87: This function `str_contains` is from PHP 8, are we dropping support for lower versions?

src/Adyen/Service.php
```diff
@@ -43,11 +43,51 @@ public function getClient()
...
+                $url = substr_replace($url, $config->get('prefix') . '-', 8, 0);
```
michaelpaul, line 88: This line contains some magic constants like "0" and "8". What about using a placeholder/variable in the url?

src/Adyen/Service.php
```diff
@@ -43,11 +43,51 @@ public function getClient()
...
     }
+
+    /**
+     * @param string $url
+     * @return string
+     */
+    protected function createBaseUrl(string $url): string
+    {
+        $config = $this->getClient()->getConfig();
+        if ($this->getClient()->getConfig()->get('enableLive')) {
```
michaelpaul, line 82: I would rather to avoid adding this new `enableLive` config. 

```suggestion
        if ($this->getClient()->getConfig()->getEnvironment() === \Adyen\Environment::LIVE) {
```

src/Adyen/HttpClient/CurlClient.php
```diff
@@ -453,4 +453,116 @@ public function requestHttp(\Adyen\Service $service, $requestUrl, $params, $meth
 
         return $result;
     }
+
+    // This method implements Restfull requests where the query params are stored in requestOptions as an array.
+    public function requestHttpRest(\Adyen\Service $service, $requestUrl, $bodyParams, $method, $requestOptions = null)
```
michaelpaul, line 458: Please add type declaration for all the parameters and output.

src/Adyen/Service.php
jillingk: No you're right not yet, replaced it with strpos()

src/Adyen/Service.php
jillingk: Is that not a bit backwards since we still have to put that placeholder there somehow with a function like that? I can just add some documentation about how it works though.

wboereboom (approved):

wboereboom:
Still see some alerts in Codacy and SonarCloud that might need to be addressed

AlexandrosMor (dismissed):

wboereboom (dismissed):

tests/Unit/ServiceTest.php
```diff
@@ -0,0 +1,55 @@
...
+        self::assertEquals("https://[prefix]-checkout-live.adyenpayments.com/checkout/[version]/[method]", $url);
+    }
+
+    public function testLiveEndpointWithoutRequiredPrefix()
+    {
+        $client = new \Adyen\Client();
+        $client->setEnvironment(Environment::LIVE);
+        $service = new Service($client);
+        try {
+            $url = $service->createBaseUrl("https://checkout-test.adyen.com/[version]/[method]");
```
wboereboom, line 38: This does not need to be a variable (throws unused variable linting warning)

src/Adyen/HttpClient/CurlClient.php
```diff
@@ -6,6 +6,11 @@
 
 class CurlClient implements ClientInterface
 {
+
+    private $UserAgent = 'User-Agent: ';
```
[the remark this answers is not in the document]
jillingk: yeah sonarcloud was complaining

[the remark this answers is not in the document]
jillingk: const doesnt work here, I suppose bc it implements a interface?

AlexandrosMor (dismissed):

wboereboom (approved):

michaelpaul (approved):

PR #458, opened 2023-03-22
[PW-8298] Added ach to ManualCapture.php to allow for manual Invoicing

carlosperales95:
**Description**
In order for ACH manual captures to work, I added the variant to the Manual capture list.

**Tested scenarios**
1. Order without ACH present in the list:
With current list, ACH payments will go to Processing automatically when the payment is fully authorised, even with `Capture Delay = Manual` setting in Magento.

2. Order with ACH present in the list:
With the new list, if capture is set to manual, the order will stay on Pending status until the CAPTURE Notification is processed, just as expected with manual capture flow.


carlosperales95:
PW was changed to ITT type, new ID is [ITT-400]

carlosperales95:
It's been a month and still reviews are required. Will this ever be merged?

jillingk:
@peterojo @candemiralp is this something you guys still need for magento? If so can you approve please.

carlosperales95:
In Magento this list has been moved to the plugin code, so it's not needed within the library (already updated that list). However it might be still useful for merchants that build their own integration using this library.

AlexandrosMor (approved):

PR #463, opened 2023-03-29
[IT-406] Generated Model Based Checkout V70

jillingk:
**Description**
This PR generated Checkout V70 with support for models. It also modifies the templates somewhat so the OneOf models don't do unnecessary allowed variable checks. 

**Tested scenarios**
Unit tests for services and UT for oneOf models: action and paymentMethod


michaelpaul (approved):
Good job! 🚀

wboereboom (approved):

Issue #462, opened 2023-03-29
Delete stored payment returns as error


gromez:
Calling `deleteStoredPaymentMethods` return a `ConnectionException` with message `Probably your Web Service username and/or password is incorrect`.

https://github.com/Adyen/adyen-php-api-library/blob/bcb1198ebce18e20abc94adcc16bb31c613dc293/src/Adyen/HttpClient/CurlClient.php#L154-L160

Maybe documentation about is wrong because I'm not getting a 200 HTTP status, but 204 with empty content, and the stored payment has been successfully deleted. 
But, because result is empty, it raises error in `handleCurlError`.

Documentation reference: https://docs.adyen.com/api-explorer/Checkout/70/delete/storedPaymentMethods/_recurringId_

PR #464, opened 2023-03-30
CurlClient request fix

jillingk:
**Description**
This PR prevents a small bug in the client that throws a curl error even though the request has been successful. This happens because an if statement catches an empty response, which is often the case with a Delete request. The fix adds an extra statement checking again if the result code is 200-204, to prevent this from happening.

**Fixed issue**: #462


src/Adyen/HttpClient/CurlClient.php
```diff
@@ -155,7 +155,7 @@ public function requestPost(\Adyen\Service $service, $requestUrl, $params)
 
         if (!in_array($httpStatus, $resultOKHttpStatusCodes) && $result) {
```
wboereboom, line 156: This code block is repeated 3 times in this class. can we generalise this to reduce code duplication?

wboereboom (changes requested):

wboereboom (approved):

AlexandrosMor (approved):

PR #465, opened 2023-03-30
[ITT-433] Fix paymentLinks endpoint URL builder

candemiralp:
**Description**

Checkout API `paymentLinks` endpoint fails due to wrong endpoint URL.

- `paymentLinks` endpoint fixed.
- Management API endpoint updated to follow same convenience

### DIFF
--- a/src/Adyen/Client.php
+++ b/src/Adyen/Client.php
@@ -33,8 +33,8 @@ class Client
     const ENDPOINT_TERMINAL_CLOUD_LIVE = "https://terminal-api-live.adyen.com";
     const ENDPOINT_TERMINAL_CLOUD_US_LIVE = "https://terminal-api-live-us.adyen.com";
     const ENDPOINT_TERMINAL_CLOUD_AU_LIVE = "https://terminal-api-live-au.adyen.com";
-    const ENDPOINT_CHECKOUT_TEST = "https://checkout-test.adyen.com/checkout";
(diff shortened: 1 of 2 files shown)


src/Adyen/Client.php
```diff
@@ -33,8 +33,8 @@ class Client
     const ENDPOINT_TERMINAL_CLOUD_LIVE = "https://terminal-api-live.adyen.com";
     const ENDPOINT_TERMINAL_CLOUD_US_LIVE = "https://terminal-api-live-us.adyen.com";
     const ENDPOINT_TERMINAL_CLOUD_AU_LIVE = "https://terminal-api-live-au.adyen.com";
-    const ENDPOINT_CHECKOUT_TEST = "https://checkout-test.adyen.com/checkout";
-    const ENDPOINT_CHECKOUT_LIVE_SUFFIX = "-checkout-live.adyenpayments.com/checkout";
```
jillingk, line 37: https://{PREFIX}-checkout-live.adyenpayments.com/checkout/v70/payments -> for live we do want the /checkout part

jillingk:
@candemiralp Can you maybe direct this PR to the main branch so we can immediately release this as a hotfix?

src/Adyen/Client.php
candemiralp: It's done!

candemiralp:
> @candemiralp Can you maybe direct this PR to the main branch so we can immediately release this as a hotfix?

I don't it is the correct approach. I would recommend creating a separate PR for release for visibility.

PR #467, opened 2023-03-30
[ITT-407] Generate Classic Payments

jillingk:
**Description**
This PR generates model based classic payments and unit tests for said classes.

**Tested scenarios**
UT's


michaelpaul (dismissed):

src/Adyen/Service/Payments/GeneralApi.php
```diff
@@ -0,0 +1,119 @@
...
... If using an API key, add an `X-API-Key` header with the API key as the value, for example:   ``` curl -H \"Content-Type: application/json\" \\ -H \"X-API-Key: YOUR_API_KEY\" \\ ... ```  Alternatively, you can use the username and password to connect to the API using basic authentication, for example:  ``` curl -U \"ws@Company.YOUR_COMPANY_ACCOUNT\":\"YOUR_BASIC_AUTHENTICATION_PASSWORD\" \\ -H \"Content-Type: application/json\" \\ ... ```  ## Versioning Payments API supports [versioning](https://docs.adyen.com/development-resources/versioning) using a version suffix in the endpoint URL. This suffix has the following format: \"vXX\", where XX is the version number.  For example: ``` https://pal-test.adyen.com/pal/servlet/Payment/v68/authorise ```  ## Going live  To authenticate to the live endpoints, you need an [API credential](https://docs.adyen.com/development-resources/api-credentials) from your live Customer Area.  The live endpoint URLs contain a prefix which is unique to your company account: ```  https://{PREFIX}-pal-live.adyenpayments.com/pal/servlet/Payment/v68/authorise ```  Get your `{PREFIX}` from your live Customer Area under **Developers** > **API URLs** > **Prefix**.
```
michaelpaul, line 5: This long description is not very useful here and repeats on every file, could you remove it please?

jillingk: Fair point no one will read it there.. how about the constructor maybe? Or just delete that summary all together?

michaelpaul (approved):

michaelpaul (approved):

PR #470, opened 2023-03-31
[ITT-402] BalancePlatform API

jillingk:
**Description**
This PR adds the BalancePlatform endpoints. Besides that it also implements a minor refactor of the curlclient (basically so we can check the url that the service classes create and check the query params are actually implemented well).

**Tested scenarios**
UT


src/Adyen/HttpClient/ClientInterface.php
```diff
@@ -33,4 +33,19 @@ public function requestPost(\Adyen\Service $service, $requestUrl, $params);
...
     public function requestHttp(\Adyen\Service $service, $requestUrl, $params, $method, $requestOptions = null);
+
+    /**
+     * @param \Adyen\Service $service
+     * @param $requestUrl
+     * @param $bodyParams
+     * @param $method
+     * @return mixed
+     */
+    public function requestHttpRest(
```
michaelpaul, line 44: Why does this service need a new public interface method? Can't it rely on existing features of the library like Management API did?

jillingk: Because of this line:  https://github.com/Adyen/adyen-php-api-library/blob/bcb1198ebce18e20abc94adcc16bb31c613dc293/src/Adyen/HttpClient/CurlClient.php#L369

Basically this prevents us from having queryparams for Post/Patch and having request body for delete for example. So I just opted for another function that splits the query params from the bodyparams.

src/Adyen/Service.php
```diff
@@ -67,7 +67,15 @@ protected function requestHttp(
             throw new AdyenException($msg);
         }
         $curlClient = $this->getClient()->getHttpClient();
-        return $curlClient->requestHttpRest($this, $url, $bodyParams, $method, $requestOptions);
+
+        // Create query params and retrieve idempotency
+        $idempotencyKey = $requestOptions['idempotencyKey'] ?? null;
+        unset($requestOptions['idempotencyKey']);
+        if (!empty($requestOptions)) {
+            $url .= '?' . http_build_query($requestOptions);
```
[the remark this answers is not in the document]
jillingk: Yep. I hadn't even thought of the possibility of more params, we'd need to change that function in the curlclient again which is a bit tedious.. How about I add the idempotency back into an array (smth like headerparams?) and pass that into the curclient?

AlexandrosMor (approved):

PR #474, opened 2023-04-04
[ITT-408] Generate Recurring API

jillingk:
**Description**
This PR adds the generated model based Recurring API + Unit Tests


tests/Unit/ModelBasedRecurringTest.php
```diff
@@ -0,0 +1,79 @@
+<?php
```
wboereboom, line 1: can we get rid of the ASCII?

wboereboom (changes requested):
We can lose the ASCII header. Looks good otherwise

michaelpaul (approved):

wboereboom (approved):

PR #501, opened 2023-05-01
[ITT-517] Fix automation pipeline

jillingk:
**Description**
This PR fixes the automation pipeline.


src/Adyen/Service/BalancePlatform/BankAccountValidationApi.php
```diff
@@ -47,10 +47,9 @@ public function __construct(Client $client)
     * @return object
     * @throws AdyenException
     */
-    public function validateBankAccountIdentification(\Adyen\Model\BalancePlatform\BankAccountIdentificationValidationRequest $bankAccountIdentificationValidationRequest, array $requestOptions = null): object
+    public function validateBankAccountIdentification(\Adyen\Model\BalancePlatform\BankAccountIdentificationValidationRequest $bankAccountIdentificationValidationRequest, array $requestOptions = null): void
```
AlexandrosMor, line 50: A validation method should return a boolean at least. In this case what is happening with this call ?

wboereboom (approved):

src/Adyen/Service/BalancePlatform/BankAccountValidationApi.php
jillingk: Actually it really returns void as apparently the endpoint returns only 200 for a successful validation.. [explorer](https://docs.adyen.com/api-explorer/balanceplatform/2/post/validateBankAccountIdentification#request-accountIdentification-SELocalAccountIdentification)
And [here](https://github.com/Adyen/adyen-openapi/blob/2f8c08266f2bdca138ba99b87eb2752a17e50171/json/BalancePlatformService-v2.json#L4095) void object in the spec. (Actual Void object btw that breaks the generators but docs is looking into that)

AlexandrosMor: This one is correct, if the validation fails it is not returning http code 200

PR #514, opened 2023-06-28
Fix multiple calls on same method with same instance

Korbeil:
**Description**
When doing the same call twice, the old updated endpoint is stored in an instance variable which cause issues.

**Tested scenarios**
```php
$checkout = new Checkout();
$result = $checkout->paymentMethods(/** args */);

// later ...

$result = $checkout->paymentMethods(/** other args */);
```
This last call will always use the same args as the first call because of how the path parameters are being updated and stored.

**Fixed issue**:  N/A

### DIFF
--- a/src/Adyen/Service/AbstractResource.php
+++ b/src/Adyen/Service/AbstractResource.php
@@ -89,8 +89,7 @@ abstract class AbstractResource
         }
 
         $params = $this->addDefaultParametersToRequest($params);
-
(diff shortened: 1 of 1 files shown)


jillingk (approved):

michaelpaul (approved):

michaelpaul:
Good catch @Korbeil and thanks for submitting a fix!

PR #518, opened 2023-07-10
Unit test for double path param substitution

jillingk:
**Description**
Quick unit test for #514.

### DIFF
--- a/tests/Unit/CheckoutTest.php
+++ b/tests/Unit/CheckoutTest.php
@@ -25,6 +25,8 @@ namespace Adyen\Tests\Unit;
 
 use Adyen\AdyenException;
 use Adyen\Service\Checkout;
+use function PHPUnit\Framework\assertContains;
+use function PHPUnit\Framework\assertEquals;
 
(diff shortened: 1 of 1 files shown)


tests/Unit/CheckoutTest.php
```diff
@@ -583,4 +585,13 @@ public static function successDeleteStoredPaymentMethodsProvider()
             array('tests/Resources/Checkout/deleteStoredPaymentMethods-success.json', 200),
         );
     }
+
+    public function testDoublePathParamSubstitution()
+    {
+        $client = $this->createMockClientUrl('tests/Resources/ModelBasedCheckout/sessions-success.json');
+        $resource = new Checkout($client);
+        $resource->captures(array('paymentPspReference' => 'pspRef1'));
```
wboereboom, line 593: I don't think we need to capture the first call in a var. this is a little bit cleaner.
```suggestion
        captures(array('paymentPspReference' => 'pspRef1'));
```

michaelpaul (approved):

wboereboom (approved):

PR #534, opened 2023-08-02
move deprecation annotations to class level

wboereboom:
**Description**
move deprecation annotations to class level so more IDE's will handle and show deprecation correctly.

### DIFF
--- a/src/Adyen/Service/Account.php
+++ b/src/Adyen/Service/Account.php
@@ -4,6 +4,7 @@ namespace Adyen\Service;
 
 /**
  * @deprecated Please consider using the BalancePlatform services instead
+ * @see Adyen\Service\BalancePlatform\AccountHoldersApi
  */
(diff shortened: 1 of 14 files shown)


wboereboom:
@peterojo Does this solve your issue?

jillingk:
I think it might be nice to use `* {@link newMethod()}` in the annotation to link to the new methods?

jillingk:
> How about [@see](https://pear.php.net/reference/PhpDocumentor-latest/phpDocumentor/tutorial_tags.see.pkg.html)?

I think link is better as `@see` shows up as "see also" in my idea which I guess is better for related classes/methods. If you use link in the deprecation message they will see deprecated with a link next to it to the class they should use instead

michaelpaul (approved):

jillingk (approved):

PR #647, opened 2024-04-05
updated API_CHECKOUT_VERSION v70 to v71

DjoykeAbyah:
bumped API_CHECKOUT_VERSION v70 to v71

**Fixed issue**:


wboereboom (approved):

wboereboom (approved):

jillingk:
@DjoykeAbyah this checkout version is for the deprecated checkout services in the library. Are we sure we want to still support this, upgrade and in doing so introduce a breaking change?

PR #679, opened 2024-07-19
chore: Fix usage of validateHMAC

cjobeilicampings:
**Description**
Hi,

We are currently working on an Adyen integration and struggled with the validateHMAC method. It looks like there were some problems:

The hmacSign sent in the webhook request is never in hexadecimal, so ctype_xdigit is always false. We must validate the HMAC key $hmacKey and use it in the pack function.

Finally, we need to compare the expectedSign with the hmacSign passed in the parameters.

Is this correct for you?

Thank you,

**Tested scenarios**

**Fixed issue**:


DjoykeAbyah:
Hi @cjobeilicampings!

Thanks for contributing a fix to this issue! We're merging this to implement your changes.

Kind Regards,
Djoyke
Adyen

DjoykeAbyah (approved):

jillingk (approved):

PR #685, opened 2024-07-29
deprecated old HmacSignature method, adjusted unittest

DjoykeAbyah:
- Deprecated old Hmac validation method 
- updated unit test with new validateHMACSignature method.

Tested by running the unit tests

### DIFF
--- a/src/Adyen/Util/HmacSignature.php
+++ b/src/Adyen/Util/HmacSignature.php
@@ -18,16 +18,16 @@ class HmacSignature
      */
     public function validateHMAC(string $hmacKey, string $hmacSign, string $webhook): bool
     {
-        if (!ctype_xdigit($hmacKey)) {
+        if (!ctype_xdigit($hmacSign)) {
             throw new AdyenException("Invalid HMAC key: $hmacKey");
         }
         $expectedSign = base64_encode(hash_hmac(
             'sha256',
             $webhook,
-            pack("H*", $hmacKey),
+            pack("H*", $hmacSign),
             true
         ));
-        return hash_equals($expectedSign, $hmacSign);
+        return hash_equals($expectedSign, $hmacKey);
     }
 
     /**
--- a/src/Adyen/Util/HmacSignature.php
+++ b/src/Adyen/Util/HmacSignature.php
@@ -9,7 +9,7 @@ class HmacSignature
     const EVENT_CODE = "eventCode";
 
     /**
-     * @deprecated Use validateHMACSignature with correct parameter order instead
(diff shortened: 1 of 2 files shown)


src/Adyen/Util/HmacSignature.php
```diff
@@ -9,6 +9,7 @@ class HmacSignature
     const EVENT_CODE = "eventCode";
 
     /**
+     * @deprecated Use validateHMACVSignature with correct parameter order instead
```
jillingk, line 12:
```suggestion
     * @deprecated Use validateHMACSignature with correct parameter order instead
```
small type

src/Adyen/Util/HmacSignature.php
```diff
@@ -28,6 +29,26 @@ public function validateHMAC(string $hmacKey, string $hmacSign, string $webhook)
...
         return hash_equals($expectedSign, $hmacKey);
     }
+    /**
+     * @param string $hmacKey Can be found in Customer Area
+     * @param string $hmacSign Can be found in the Webhook headers
+     * @param string $webhook The response from Adyen
+     * @return bool
+     * @throws AdyenException
+     */
+    public function validateHMACSignature(string $hmacKey, string $hmacSign, string $webhook): bool
```
jillingk, line 39: Are you not switching it up now? The code just committed should be the way forward and this method is the "old" wrong way no?

DjoykeAbyah: Ah you're right it seems like I've pulled the non updated version!

jillingk: Ahhh okay, you got it right. Thought you already rebased on the contribution but you didn't :)

jillingk (approved):

tests/Unit/Util/HmacSignatureTest.php
```diff
@@ -164,7 +164,7 @@ public function testBankingWebhookHmacValidation()
             . "{\"balanceAccountId\":\"BA3227C223222H5HVKT3H9WLC\"},\"currency\":\"EUR\"}},\"environment\":"
             . "\"test\",\"type\":\"balancePlatform.balanceAccountSweep.updated\"}";
         $hmac = new HmacSignature();
-        $result = $hmac->validateHMAC(
+        $result = $hmac->validateHMACSignature(
```
acampos1916, line 167: Is it possible to keep both unit tests while the method is being deprecated? And then we remove the unit test once fully gone

DjoykeAbyah: Yes ofcourse I'll commit with both tests

jillingk (approved):

PR #717, opened 2024-12-09
added APSE to VALID_REGIONS

DjoykeAbyah:
Added APSE to valid regions since we have a live endpoint added to TERMINAL_API_ENDPOINTS_MAPPING

### DIFF
--- a/tests/Unit/RegionTest.php
+++ b/tests/Unit/RegionTest.php
@@ -30,7 +30,6 @@ class RegionTest extends TestCase
             "eu" => "https://terminal-api-live.adyen.com",
             "us" => "https://terminal-api-live-us.adyen.com",
             "au" => "https://terminal-api-live-au.adyen.com",
-            "in" => "https://terminal-api-live-in.adyen.com",
             "apse" => "https://terminal-api-live-apse.adyen.com",
         ];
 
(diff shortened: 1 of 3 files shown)


Kwok-he-Chu (changes requested):
Let's add the tests that allow us to spot these changes from now on if we decide to modify it:
- Test for the VALID_REGIONS
- Test for each TERMINAL_API_ENDPOINTS_MAPPING (and whether they are mapping correctly

DjoykeAbyah:
@Kwok-he-Chu,  I've included the unit test. Since India is also part of valid regions I've included the url in Client.php and added it to the mapping.

ayodejidev (dismissed):
Looks good!

tests/Unit/RegionTest.php
```diff
@@ -0,0 +1,56 @@
...
+        );
+    }
+
+    public function testTerminalApiEndpointsMapping()
+    {
+        $expected = [
+            "eu" => "https://terminal-api-live.adyen.com",
+            "us" => "https://terminal-api-live-us.adyen.com",
+            "au" => "https://terminal-api-live-au.adyen.com",
+            "in" => "https://terminal-api-live-in.adyen.com",
```
Kwok-he-Chu, line 33: Could you verify whether the`live-in` endpoint is a valid for terminals

Kwok-he-Chu: It's not a valid one, let's remove this one and update the tests https://docs.adyen.com/point-of-sale/design-your-integration/terminal-api/ @DjoykeAbyah

tests/Unit/RegionTest.php
```diff
@@ -0,0 +1,56 @@
...
+        ];
+
+        $this->assertEquals(
+            $expected,
+            Region::TERMINAL_API_ENDPOINTS_MAPPING,
+            "TERMINAL_API_ENDPOINTS_MAPPING should match the expected mappings."
+        );
+    }
+
+    public function testTerminalApiEndpointsExistsForAllRegions()
```
Kwok-he-Chu, line 44: We can remove this test as not every region (as of currently*) has an endpoint yet (e.g. india) and this test will fail -> see https://docs.adyen.com/point-of-sale/design-your-integration/terminal-api/

At some point, we will  (once we're live)

tests/Unit/RegionTest.php
DjoykeAbyah: Done!

Issue #710, opened 2024-11-18
GuzzleHttp\Utils is not imported


szepeviktor:
**Describe the bug**
`GuzzleHttp\Utils` is not imported but used

**To Reproduce**
https://github.com/search?q=repo%3AAdyen%2Fadyen-php-api-library%20%22Utils%22&type=code

**Expected behavior**
`use GuzzleHttp\Utils;`

szepeviktor:
Please consider using @phpstan

szepeviktor:
Remnants from https://github.com/Adyen/adyen-php-api-library/pull/449#discussion_r1128125526

gcatanese:
Fixed by #720

PR #720, opened 2024-12-12
Import Guzzle utils

szepeviktor:
**Description**
`GuzzleHttp\Utils` is not imported but used

**Tested scenarios**
Running @phpstan

**Fixed issue**:  #710


szepeviktor:
Maybe we should remove this line
https://github.com/Adyen/adyen-php-api-library/blob/af262a702228c473bbb4c00f9d7000f0d85ad7e5/templates/ObjectSerializer.mustache#L227

DjoykeAbyah:
Hi @szepeviktor! 

Thank you for your contribution and for pointing this out. We're reviewing it now and will get back to you shortly.

Kind regards,

Djoyke
Adyen

szepeviktor:
I think Guzzle was removed long time ago, and this is the last remnant.

gcatanese (approved):

PR #748, opened 2025-03-10
Add nullable return types to API mustache templates

ayodejidev:
**Description**

Updated API mustache tempates to include nullable return types to better reflect their behavior, improve type safety and support php v8+.


gcatanese (changes requested):
Looks good.

I suggest to remove `NullableTypeDeclarationTest` and instead consider adding a unit test in the `SDK Automation` generation to verify the signature of the generated methods.

gcatanese (approved):

PR #749, opened 2025-03-10
Remove implicit nullable types

Nek-:
**Description**

This should remove a lot of deprecation notice while running the library with php 8.4.

I used rector with the following configurations:

```php
<?php
// rector.php
return Rector\Config\RectorConfig::configure()
    ->withPaths([__DIR__ . '/src'])
    ->withPhpVersion(Rector\ValueObject\PhpVersion::PHP_84)
    ->withRules([
         Rector\Php84\Rector\Param\ExplicitNullableParamTypeRector::class,
    ]);
```

Since the code is generated I suggest you to add this in your CI/CD flow. Once the config file is added to the project, just run `cpx rector/rector`.


aherrao:
@Nek- Thank you so much for raising this issue.
Can you please do the needful? We are also looking for these changes in our project update.

Nek-:
I think it's done in https://github.com/Adyen/adyen-php-api-library/pull/748

aherrao:
@Nek- Thanks for this information.

PR #881, opened 2026-02-02
[management] Code generation: update services and models

AdyenAutomationBot:
This PR contains the automated changes for the `management` service.

The commit history of this PR reflects the `adyen-openapi` commits that have been applied.


andrej-gajduk-adyen:
Breaking changes: YES

- OfflineProcessing.php: Removed public methods getOfflineSwipeLimits() and setOfflineSwipeLimits(), and the offlineSwipeLimits property was entirely dropped from the model.
- AffirmInfo.php: Removed public constants PRICE_PLAN_BRONZE, PRICE_PLAN_SILVER, PRICE_PLAN_GOLD and the public method getPricePlanAllowableValues(). Enum validation in setPricePlan() was also removed; the field is now treated as a free-form string.

Suspicious patterns: YES

- AffirmInfo.pricePlan: The updated docstring describes three new valid values ("core", "standard", "signature"), but no enum constants were generated for them. Other fields in this same PR with enumerated values (e.g., AccelResponseInfo.processingType) do receive enum constants. The inconsistency suggests the generator may have treated this field as a plain string when the spec still intends it to be an enum.
- CreateCompanyUserRequest and CreateMerchantUserRequest: The loginMethod docstring had "Username & account" silently removed as a valid option, with no corresponding constant or validation change in code -- callers relying on this value would receive no error at the PHP level.

andrej-gajduk-adyen:
marking as breaking change because of removal of field `OfflineSwipeLimits` in `OfflineProcessing`
and change to enum values for price plan `AffirmInfo`

andrej-gajduk-adyen (approved):
Breaking change:
- removal of field OfflineSwipeLimits in OfflineProcessing
- change to enum values for price plan AffirmInfo

PR #887, opened 2026-02-16
[transfers] Code generation: update services and models

AdyenAutomationBot:
This PR contains the automated changes for the `transfers` service.

The commit history of this PR reflects the `adyen-openapi` commits that have been applied.


andrej-gajduk-adyen:
Breaking changes: YES

- CapitalApi service class renamed to CapitalDeprecatedApi (src/Adyen/Service/Transfers/CapitalApi.php → CapitalDeprecatedApi.php); any code instantiating new CapitalApi(...) or type-hinting against it will break.
- Counterparty model class removed/renamed to GrantCounterparty (src/Adyen/Model/Transfers/Counterparty.php → GrantCounterparty.php); direct use of Counterparty class will fail.
- CapitalGrant::getCounterparty() / setCounterparty() return/parameter type changed from Counterparty to GrantCounterparty.
- CapitalGrantInfo::getCounterparty() / setCounterparty() return/parameter type changed from Counterparty to GrantInfoCounterparty.

Suspicious patterns: YES

- TransferEventEventsDataInner now has DISCRIMINATOR = 'type' and auto-initializes type to static::$openAPIModelName (= 'TransferEventEventsDataInner'), but this class is a merged/flattened OneOf combining interchange and capture data fields. The discriminator value 'TransferEventEventsDataInner' will never match a real API type string, so discriminator-based deserialization from this class will silently set incorrect type values when instantiated directly.
- TransferCategoryData and TransferDataTracking / TransferEventTrackingData have the same discriminator self-initialization pattern (type set to class name, not API value); this is consistent across all modified discriminator classes and may be a generator-level issue.
- BankAccountV3AccountIdentification.php gains DISCRIMINATOR = 'type' with type auto-set to 'BankAccountV3AccountIdentification', which is unlikely to be the actual API wire value for any concrete account identification subtype.

andrej-gajduk-adyen:
Marking as breaking change due to renames of `CapitalApi` and `Counterparty`

andrej-gajduk-adyen (approved):
Breaking change:
- CapitalApi renamed to CapitalDeprecatedApi
- Counterparty renamed to GrantCounterparty

PR #897, opened 2026-03-11
1. OpenAPI generator upgrade: add templates and shared classes

gcatanese:
First iteration of the OpenAPI Generator upgrade:

- add and customise Mustache templates
- add code shared by different APIs (`Configuration`, `BaseService`, etc...)
- generate `BinlookUp` services and models
- add tests for shared code and `BinlookUp` to verify main use cases: initialise objects, send requests, add custom headers, throw exceptions upon `4xx`/`5xx` errors

At the moment the new code co-exist with the old version, therefore there is a bit of mix of approaches and code, I also had to disable few tests (to verify/refactor later). This will improve as we progress this migration.


src/Adyen/Service/BinLookup/BinLookupApi.php
```diff
@@ -0,0 +1,902 @@
...
+        if (isset($decodedPayload['message']) && isset($decodedPayload['errorCode'])) {
+            return new \Adyen\AdyenException(
+                $decodedPayload['message'],
+                $decodedPayload['status'] ?? 0,
+                null,
+                $decodedPayload['status'] ?? null,
+                $decodedPayload['errorType'] ?? null,
+                $decodedPayload['pspReference'] ?? null,
+                $decodedPayload['errorCode']
+            );
```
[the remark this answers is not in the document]
gcatanese: You are looking at the wrong class,`Adyen\AdyenException` constructor has 7 attributes

src/Adyen/BaseService.php
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * Service constructor.
+     *
+     * @param Configuration $configuration
+     * @throws AdyenException
+     */
+    public function __construct(Configuration $configuration)
+    {
+        if (!$configuration->getAdyenApiKey()) {
```
ashwaniarya-adyen, line 20: Are we making this change intentional ?
Because in the older version, we have
`elseif ($service->requiresApiKey()) 
{ $msg = "Please provide a valid Checkout API Key"; 
throw new AdyenException($msg);`
in client where we depend on the flag passed by service to determine wether we need to validate API key

templates-v7/api.mustache
```diff
@@ -0,0 +1,914 @@
...
+{{/servers}}
+     * @param  \{{invokerPackage}}\RequestOptions|null $requestOptions Additional request options (optional)
+     *
+     * @throws \InvalidArgumentException
+     * @return \GuzzleHttp\Promise\PromiseInterface
+    {{#isDeprecated}}
+     * @deprecated
+    {{/isDeprecated}}
+     */
+    public function {{operationId}}AsyncWithHttpInfo({{^exts.x-group-parameters}}{{#allParams}}{{dataType}} ${{paramName}}{{^isBodyParam}}{{^required}} = {{#defaultValue}}{{{.}}}{{/defaultValue}}{{^defaultValue}}null{{/defaultValue}}{{/required}}{{/isBodyParam}}, {{/allParams}}{{#servers}}{{#-first}}?int $hostIndex = null, array $variables = [], {{/-first}}{{/servers}}\{{invokerPackage}}\RequestOptions $requestOptions = null{{/exts.x-group-parameters}}{{#exts.x-group-parameters}}$associative_array{{/exts.x-group-parameters}}): \GuzzleHttp\Promise\PromiseInterface
```
ashwaniarya-adyen: The async success callback does not check the response status, so a 401/500 is deserialized and returned as a successful response model.  
Could we reject non-2xx responses here, matching the synchronous methods?

src/Adyen/Exception/AdyenException.php
```diff
@@ -0,0 +1,112 @@
...
+namespace Adyen\Exception;
+
+use Exception;
+
+/**
+ * ApiException Class
+ *
+ * @package  Adyen
+ */
+class AdyenException extends Exception
```
ashwaniarya-adyen: I think we can rename this exception class as HTTPException, since it's responsibility is dealing with HTTP Response
Also, src/Adyen/AdyenException.php, can be moved in src/Adyen/Exception for cleaner view

templates-v7/model_generic.mustache
```diff
@@ -0,0 +1,596 @@
...
+            $nullablesSetToNull = $this->getOpenAPINullablesSetToNull();
+            $index = array_search('{{name}}', $nullablesSetToNull);
+            if ($index !== FALSE) {
+                unset($nullablesSetToNull[$index]);
+                $this->setOpenAPINullablesSetToNull($nullablesSetToNull);
+            }
+        }
+        {{/isNullable}}
+        {{^isNullable}}
+        if (is_null(${{name}})) {
```
ashwaniarya-adyen, line 401: Is this stricter setter behavior an intentional breaking SDK change? Previously, setters accepted `null`, and the SDK did not automatically call `valid()` before sending a request

composer.json
```diff
@@ -8,19 +8,18 @@
   "homepage": "https://github.com/Adyen/adyen-php-api-library",
   "license": "MIT",
   "require": {
-    "php": ">=7.3",
+    "php": "^8.1",
```
ashwaniarya-adyen, line 11: Update Composer requirements for the v7 upgrade, requires PHP 8.1, Guzzle, and PSR-7, align PHPUnit, PHP-CS-Fixer, and PHPCS version constraints.

.github/workflows/main.yml
```diff
@@ -13,15 +13,15 @@ on:
...
+  contents: read
 
 jobs:
   php-test:
     name: PHP Test
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php-version: [ '7.3', '8.2', '8.4', '8.5' ]
+        php-version: [ '8.1', '8.2', '8.4' ]
```
ashwaniarya-adyen, line 24: Adds PHP 8.1 to the test matrix because it is the minimum PHP version declared by this branch.

src/Adyen/Configuration.php
```diff
@@ -0,0 +1,655 @@
...
+    }
+
+    /**
+     * Returns URL based on the index and variables
+     *
+     * @param int $index index of the host settings
+     * @param array|null $variables hash of variable and the corresponding value (optional)
+     * @return string|null URL based on host settings
+     */
+    public function getHostFromSettings(int $index, ?array $variables = null): ?string
```
ashwaniarya-adyen, line 651: Declare the optional host variables argument as `?array`. PHP 8.4+ deprecates `array $variables = null`, and main converts deprecations into PHPUnit errors because of the change [here](https://github.com/Adyen/adyen-php-api-library/blob/aeb3d01eede213f09109f197b5487dc80dcaee1a/phpunit.xml#L28)

src/Adyen/Service/BinLookup/BinLookupApi.php
```diff
@@ -0,0 +1,902 @@
...
+     * @param int $hostIndex (Optional) host index to select the list of hosts if defined in the OpenAPI spec
+     * @throws \Adyen\AdyenException
+     */
+    public function __construct(
+        ?Configuration $config = null,
+        ?ClientInterface $client = null,
+        ?HeaderSelector $selector = null,
+        int $hostIndex = 0
+    ) {
+        $this->config = $config ?: Configuration::getDefaultConfiguration();
```
ashwaniarya-adyen, line 90: Resolve the default configuration before calling `BaseService`. Previously, `new BinLookupApi()` passed `null` to `BaseService`, causing a PHP `TypeError` before the SDK could use its default configuration or return its normal configuration error.

templates-v7/api.mustache
```diff
@@ -0,0 +1,914 @@
...
+     * @param int $hostIndex (Optional) host index to select the list of hosts if defined in the OpenAPI spec
+     * @throws \Adyen\AdyenException
+     */
+    public function __construct(
+        ?Configuration $config = null,
+        ?ClientInterface $client = null,
+        ?HeaderSelector $selector = null,
+        int $hostIndex = 0
+    ) {
+        $this->config = $config ?: Configuration::getDefaultConfiguration();
```
ashwaniarya-adyen, line 80: Explanation here - https://github.com/Adyen/adyen-php-api-library/pull/897/changes#r3979443868

src/Adyen/Service/BinLookup/BinLookupApi.php
```diff
@@ -0,0 +1,902 @@
...
+            $defaultHeaders['User-Agent'] = $this->config->getUserAgent();
+        }
+
+        $headers = array_merge(
+            $defaultHeaders,
+            $headerParams,
+            $headers
+        );
+
+        $operationHost = $this->config->getHost() ?: $this->baseURL;
```
ashwaniarya-adyen, line 457: When no custom host was configured, `getHost()` returned an empty value and the generated request contains only a relative path. Guzzle cannot send this because it has no Adyen server address. Fall back to the BinLookup base URL to build a complete request URL while preserving any merchant-provided custom host.

templates-v7/api.mustache
```diff
@@ -0,0 +1,914 @@
...
+
+        $hostSettings = $this->getHostSettingsFor{{operationId}}();
+
+        if ($hostIndex < 0 || $hostIndex >= count($hostSettings)) {
+            throw new \InvalidArgumentException("Invalid index {$hostIndex} when selecting the host. Must be less than ".count($hostSettings));
+        }
+        $operationHost = Configuration::getHostString($hostSettings, $hostIndex, $variables);
+        {{/servers.0}}
+        {{^servers.0}}
+        $operationHost = $this->config->getHost() ?: $this->baseURL;
```
ashwaniarya-adyen, line 766: Explanation here - https://github.com/Adyen/adyen-php-api-library/pull/897/changes#r3979488839

src/Adyen/Service/BinLookup/BinLookupApi.php
```diff
@@ -0,0 +1,902 @@
...
```
ashwaniarya-adyen, line 796: Explanation here - https://github.com/Adyen/adyen-php-api-library/pull/897/changes#r3979488839

src/Adyen/Service/BinLookup/BinLookupApi.php
```diff
@@ -0,0 +1,902 @@
...
+     * Check if 3D Secure is available
+     *
+     * @param  \Adyen\Model\BinLookup\ThreeDSAvailabilityRequest|null $threeDSAvailabilityRequest threeDSAvailabilityRequest (optional)
+     * @param  \Adyen\RequestOptions|null $requestOptions Additional request options (optional)
+     *
+     * @throws \Adyen\Exception\AdyenException on non-2xx response or if the response body is not in the expected format
+     * @throws \InvalidArgumentException
+     * @return \Adyen\Model\BinLookup\ThreeDSAvailabilityResponse
+     */
+    public function get3dsAvailability(\Adyen\Model\BinLookup\ThreeDSAvailabilityRequest$threeDSAvailabilityRequest, ?\Adyen\RequestOptions $requestOptions = null): \Adyen\Model\BinLookup\ThreeDSAvailabilityResponse
```
ashwaniarya-adyen: The generated methods gave `RequestOptions` a default value of `null` but declared it as a non-nullable type. PHP 8.4+ treats this implicit nullable declaration as deprecated. Added `?` to the type -> `?\Adyen\RequestOptions $requestOptions = null`, so the declared type matches the default value and no deprecation warning is raised. [Here](https://github.com/Adyen/adyen-php-api-library/pull/897/changes#r3979377529)

src/Adyen/Service/BinLookup/BinLookupApi.php
```diff
@@ -0,0 +1,902 @@
...
+     * Check if 3D Secure is available
+     *
+     * @param  \Adyen\Model\BinLookup\ThreeDSAvailabilityRequest|null $threeDSAvailabilityRequest (optional)
+     * @param  \Adyen\RequestOptions|null $requestOptions Additional request options (optional)
+     *
+     * @throws \Adyen\Exception\AdyenException on non-2xx response or if the response body is not in the expected format
+     * @throws \InvalidArgumentException
+     * @return array of \Adyen\Model\BinLookup\ThreeDSAvailabilityResponse, HTTP status code, HTTP response headers (array of strings)
+     */
+    public function get3dsAvailabilityWithHttpInfo(?\Adyen\Model\BinLookup\ThreeDSAvailabilityRequest $threeDSAvailabilityRequest = null, ?\Adyen\RequestOptions $requestOptions = null): array
```
ashwaniarya-adyen: Explanation here - https://github.com/Adyen/adyen-php-api-library/pull/897/changes#r3979553967

templates-v7/api.mustache
```diff
@@ -0,0 +1,914 @@
...
+     * @param  \{{invokerPackage}}\RequestOptions|null $requestOptions Additional request options (optional)
+     *
+     * @throws \{{invokerPackage}}\Exception\AdyenException on non-2xx response or if the response body is not in the expected format
+     * @throws \InvalidArgumentException
+     * @return {{#returnType}}{{.}}{{/returnType}}{{^returnType}}void{{/returnType}}
+    {{#isDeprecated}}
+     * @deprecated
+    {{/isDeprecated}}
+     */
+    public function {{operationId}}({{^exts.x-group-parameters}}{{#allParams}}{{dataType}}${{paramName}}{{^isBodyParam}}{{^required}} = {{#defaultValue}}{{{.}}}{{/defaultValue}}{{^defaultValue}}null{{/defaultValue}}{{/required}}{{/isBodyParam}}, {{/allParams}}{{#servers}}{{#-first}}?int $hostIndex = null, array $variables = [], {{/-first}}{{/servers}}?\{{invokerPackage}}\RequestOptions $requestOptions = null{{/exts.x-group-parameters}}{{#exts.x-group-parameters}}$associative_array{{/exts.x-group-parameters}}): {{#returnType}}{{.}}{{/returnType}}{{^returnType}}void{{/returnType}}
```
ashwaniarya-adyen: Explanation Here - https://github.com/Adyen/adyen-php-api-library/pull/897/changes#r3979553967

templates-v7/api.mustache
```diff
@@ -0,0 +1,914 @@
...
+     * @param  \{{invokerPackage}}\RequestOptions|null $requestOptions Additional request options (optional)
+     *
+     * @throws \{{invokerPackage}}\Exception\AdyenException on non-2xx response or if the response body is not in the expected format
+     * @throws \InvalidArgumentException
+     * @return array of {{#returnType}}{{.}}{{/returnType}}{{^returnType}}null{{/returnType}}, HTTP status code, HTTP response headers (array of strings)
+    {{#isDeprecated}}
+     * @deprecated
+    {{/isDeprecated}}
+     */
+    public function {{operationId}}WithHttpInfo({{^exts.x-group-parameters}}{{#allParams}}{{#notRequiredOrIsNullable}}?{{/notRequiredOrIsNullable}}{{dataType}} ${{paramName}}{{^required}} = {{#defaultValue}}{{{.}}}{{/defaultValue}}{{^defaultValue}}null{{/defaultValue}}{{/required}}, {{/allParams}}{{#servers}}{{#-first}}?int $hostIndex = null, array $variables = [], {{/-first}}{{/servers}}?\{{invokerPackage}}\RequestOptions $requestOptions = null{{/exts.x-group-parameters}}{{#exts.x-group-parameters}}$associative_array{{/exts.x-group-parameters}}): array
```
ashwaniarya-adyen: https://github.com/Adyen/adyen-php-api-library/pull/897/changes#r3979553967

templates-v7/api.mustache
```diff
@@ -0,0 +1,914 @@
...
+{{/servers}}
+     * @param  \{{invokerPackage}}\RequestOptions|null $requestOptions Additional request options (optional)
+     *
+     * @throws \InvalidArgumentException
+     * @return \GuzzleHttp\Promise\PromiseInterface
+    {{#isDeprecated}}
+     * @deprecated
+    {{/isDeprecated}}
+     */
+    public function {{operationId}}AsyncWithHttpInfo({{^exts.x-group-parameters}}{{#allParams}}{{dataType}} ${{paramName}}{{^isBodyParam}}{{^required}} = {{#defaultValue}}{{{.}}}{{/defaultValue}}{{^defaultValue}}null{{/defaultValue}}{{/required}}{{/isBodyParam}}, {{/allParams}}{{#servers}}{{#-first}}?int $hostIndex = null, array $variables = [], {{/-first}}{{/servers}}?\{{invokerPackage}}\RequestOptions $requestOptions = null{{/exts.x-group-parameters}}{{#exts.x-group-parameters}}$associative_array{{/exts.x-group-parameters}}): \GuzzleHttp\Promise\PromiseInterface
```
ashwaniarya-adyen: https://github.com/Adyen/adyen-php-api-library/pull/897/changes#r3979553967

templates-v7/api.mustache
```diff
@@ -0,0 +1,914 @@
...
+     * @param  string $contentType The value for the Content-Type header. Check self::CONTENT_TYPES['{{{operationId}}}'] to see the possible values for this operation
+     * @param \Adyen\RequestOptions|null $requestOptions
+     *
+     * @throws \InvalidArgumentException
+     * @return \GuzzleHttp\Psr7\Request
+    {{#isDeprecated}}
+     * @deprecated
+    {{/isDeprecated}}
+     */
+    public function {{operationId}}Request({{^exts.x-group-parameters}}{{#allParams}}{{{dataType}}} ${{paramName}}{{^isBodyParam}}{{^required}} = {{#defaultValue}}{{{.}}}{{/defaultValue}}{{^defaultValue}}null{{/defaultValue}}{{/required}}{{/isBodyParam}}, {{/allParams}}{{#servers}}{{#-first}}?int $hostIndex = null, array $variables = [], {{/-first}}{{/servers}}string $contentType = self::CONTENT_TYPES['{{{operationId}}}'][0]{{/exts.x-group-parameters}}{{#exts.x-group-parameters}}$associative_array{{/exts.x-group-parameters}}, ?\Adyen\RequestOptions $requestOptions = null): Request
```
ashwaniarya-adyen: https://github.com/Adyen/adyen-php-api-library/pull/897/changes#r3979553967

src/Adyen/Model/BinLookup/Amount.php
```diff
@@ -261,13 +258,19 @@ private function setIfExists(string $variableName, array $fields, $defaultValue)
...
      */
-    public function listInvalidProperties()
+    public function listInvalidProperties(): array
     {
         $invalidProperties = [];
 
         if ($this->container['currency'] === null) {
             $invalidProperties[] = "'currency' can't be null";
         }
+        if ($this->container['currency'] !== null && mb_strlen($this->container['currency']) > 3) {
```
ashwaniarya-adyen, line 268: A required property can still be `null` while a model is being validated. The validation method adds the correct “can’t be null” error but continues to the length checks. Calling `mb_strlen(null)` is deprecated in PHP 8.1+, so check that the value is not null before checking its length. [Here](https://github.com/Adyen/adyen-php-api-library/pull/897/changes#r3979377529)

templates-v7/model_generic.mustache
```diff
@@ -0,0 +1,594 @@
...
+                $this->container['{{name}}'],
+                implode("', '", $allowedValues)
+            );
+        }
+
+        {{/isContainer}}
+        {{/isEnum}}
+        {{#hasValidation}}
+        {{#maxLength}}
+        if ($this->container['{{name}}'] !== null && mb_strlen($this->container['{{name}}']) > {{maxLength}}) {
```
ashwaniarya-adyen, line 306: Explanation Here - https://github.com/Adyen/adyen-php-api-library/pull/897/changes#r3979377529

templates-v7/model_generic.mustache
```diff
@@ -0,0 +1,594 @@
...
+            $invalidProperties[] = "invalid value for '{{name}}', the character length must be smaller than or equal to {{{maxLength}}}.";
+        }
+        {{/maxLength}}
+        {{#minLength}}
+        if ($this->container['{{name}}'] !== null && mb_strlen($this->container['{{name}}']) < {{minLength}}) {
```
ashwaniarya-adyen, line 311: Explanation Here - https://github.com/Adyen/adyen-php-api-library/pull/897/changes#r3979377529

src/Adyen/Model/BinLookup/ObjectSerializer.php
```diff
@@ -15,9 +12,15 @@
 
 namespace Adyen\Model\BinLookup;
 
+use Adyen\Configuration;
```
ashwaniarya-adyen, line 15: Without this import, PHP looks for `Adyen\Model\BinLookup\Configuration`, which does not exist, instead of the SDK class `Adyen\Configuration`. Import the SDK configuration class so serializer configuration lookups resolve correctly.

templates-v7/ObjectSerializer.mustache
```diff
@@ -0,0 +1,581 @@
+<?php
+{{>partial_header}}
+
+namespace {{modelPackage}};
+
+use Adyen\Configuration;
```
ashwaniarya-adyen, line 6: Explanation here - https://github.com/Adyen/adyen-php-api-library/pull/897/changes#r3979784063

tests/Unit/BinLookupTest.php
```diff
@@ -2,33 +2,362 @@
...
+        $this->assertEquals('idempotencyKeyAsync', $request->getHeaderLine('Idempotency-Key'));
+        $this->assertEquals('CustomValueAsync', $request->getHeaderLine('Custom-Header-Async'));
+
+        $this->assertInstanceOf(ThreeDSAvailabilityResponse::class, $result);
+        $this->assertEquals(200, $statusCode);
+        $this->assertTrue($result->getThreeDs1Supported());
+        $this->assertEmpty($headers);
+    }
+
+    public function testRequestUsesBaseUrl()
```
ashwaniarya-adyen, line 335: Added a regression test for the missing-host case. It verifies that BinLookup creates a complete URL using its default base URL, rather than a relative endpoint path that Guzzle cannot send.

tests/Unit/BinLookupTest.php
```diff
@@ -2,33 +2,362 @@
...
+            new ThreeDSAvailabilityRequest()
+        );
+
+        $this->assertEquals(
+            'https://pal-test.adyen.com/pal/servlet/BinLookup/v54/get3dsAvailability',
+            (string) $request->getUri()
+        );
+    }
+
+    public function testDefaultConfiguration()
```
ashwaniarya-adyen: Added a regression test for construction without an explicit configuration. It verifies that `BinLookupApi` loads the SDK default configuration before it reaches `BaseService`, instead of passing `null` and causing a PHP type error.

src/Adyen/BaseService.php
gcatanese: Interesting. I don't think we should allow the option to skip API key validation. All our APIs are behind authentication.
I notice though that we don't have support for Basic Authentication, which is a legacy authentication but still relevant (according to the [Docs](https://docs.adyen.com/development-resources/api-credentials#basic-authentication)).
We should introduce it, either here or create a new ticket for later.

tests/Unit/BinLookupTest.php
gcatanese: Very good, maybe consider renaming to `testConstructorUsesDefaultConfiguration`

templates-v7/model_generic.mustache
gcatanese: Good point. In theory we shouldn't accept `null` unless the attribute is `nullable`, but this is a significant breaking change (as you point out 👍). The other important aspect is that we want to avoid validation in the SDKs (when possible) so we avoid drifting from the API (for example the API changes a property to be nullable but the merchants don't upgrade the SDK version).

This is the same reason why we don't call `valid()` before sending a request.

To summarise:

- remove validation of nullability (no longer throw the exception): we rely on the validation of the API
- `valid`: consider if it is better to remove or make a comment

src/Adyen/Service/BinLookup/BinLookupApi.php
gcatanese: We need to revise this since custom hosts are not a supported merchant feature. API URLs are defined in `createBaseUrl`, and they support:

- `test` and `live` host
- for payments only we support a merchant prefix i.e. `https://abcdef-pal-live.adyenpayments.com`

We should maybe cleanup the code to avoid even providing a custom host, to be investigated

templates-v7/api.mustache
gcatanese: See comment about custom host

src/Adyen/Service/BinLookup/BinLookupApi.php
gcatanese: See comment about custom host

src/Adyen/Service/BinLookup/BinLookupApi.php
gcatanese: Space missing between type and variable name in the signature, only in the first parameter
```
get3dsAvailability(\Adyen\Model\BinLookup\ThreeDSAvailabilityRequest$threeDSAvailabilityRequest,
```

It will require a template change

gcatanese: There is an inconsistency across similar methods:
- `get3dsAvailability`:  main method, body required
- `get3dsAvailabilityWithHttpInfo`: WithHttpInfo variant, body optional and nullable
- `get3dsAvailabilityRequest`: request builder, body required and non-nullable

Methods could be aligned to define the body as required and non-nullable.

Unfortunately our specs don't mark the request body as `required: true`, so we cannot rely on the OpenAPI contract. We don't have a use case of a request with an empty body (for `POST` and `PATCH`), so we can assume here that we always need one.

src/Adyen/Model/BinLookup/Amount.php
gcatanese: Good work. The null guard is applied to `maxLength`/`minLength`, but the same issue appears in the template for pattern (preg_match(pattern, null), deprecated in PHP 8.1+) and maxItems/minItems (count(null), TypeError) when a required property is null.

Note that `valid()` and `listInvalidProperties()` are never called by the SDK itself, we should assess what's best:
- remove them (code is clearer and cleaner)
- leave them optional for merchants to use (to be mentioned in the README maybe)

If decide to go for option 2, we may need to revise all validations scenario, I suggest to create a new ticket for this step.

templates-v7/model_generic.mustache
gcatanese: See comment about validation above

templates-v7/model_generic.mustache
gcatanese: See comment about validation above

gcatanese (commented):
Great work @ashwaniarya-adyen, I left some comments, we need to iterate over few points, but overall it looks really good!

templates-v7/api.mustache
```diff
@@ -0,0 +1,915 @@
...
+     * @param  \{{invokerPackage}}\RequestOptions|null $requestOptions Additional request options (optional)
+     *
+     * @throws \{{invokerPackage}}\Exception\AdyenException on non-2xx response or if the response body is not in the expected format
+     * @throws \InvalidArgumentException
+     * @return {{#returnType}}{{.}}{{/returnType}}{{^returnType}}void{{/returnType}}
+    {{#isDeprecated}}
+     * @deprecated
+    {{/isDeprecated}}
+     */
+    public function {{operationId}}({{^exts.x-group-parameters}}{{#allParams}}{{^isBodyParam}}{{#notRequiredOrIsNullable}}?{{/notRequiredOrIsNullable}}{{/isBodyParam}}{{dataType}} ${{paramName}}{{^isBodyParam}}{{^required}} = {{#defaultValue}}{{{.}}}{{/defaultValue}}{{^defaultValue}}null{{/defaultValue}}{{/required}}{{/isBodyParam}}, {{/allParams}}{{#servers}}{{#-first}}?int $hostIndex = null, array $variables = [], {{/-first}}{{/servers}}?\{{invokerPackage}}\RequestOptions $requestOptions = null{{/exts.x-group-parameters}}{{#exts.x-group-parameters}}$associative_array{{/exts.x-group-parameters}}): {{#returnType}}{{.}}{{/returnType}}{{^returnType}}void{{/returnType}}
```
ashwaniarya-adyen: Optional parameters, such as an idempotency key in checkout, can be left empty. Their type must therefore allow `null`, otherwise PHP 8.4 shows a deprecation warning. Request bodies remain required. This also fixes the missing space in the generated parameter declaration.

src/Adyen/Service/BinLookup/BinLookupApi.php
ashwaniarya-adyen: Done here - https://github.com/Adyen/adyen-php-api-library/pull/897/changes#r3988827681

ashwaniarya-adyen: Done here - https://github.com/Adyen/adyen-php-api-library/pull/897/changes#r3988827681

ashwaniarya-adyen: `get3dsAvailability` - [Link](https://github.com/Adyen/adyen-php-api-library/blob/82a321689e830721577e8566b58ee8c16a831c9c/templates-v7/api.mustache#L172)
`get3dsAvailabilityWithHttpInfo` -  [Link](https://github.com/Adyen/adyen-php-api-library/blob/82a321689e830721577e8566b58ee8c16a831c9c/templates-v7/api.mustache#L231)
`get3dsAvailabilityRequest ` - [Link](https://github.com/Adyen/adyen-php-api-library/blob/82a321689e830721577e8566b58ee8c16a831c9c/templates-v7/api.mustache#L547)

Null is generated only when the parameter is not a body parameter. Therefore, body parameters remain required and non-nullable across all three generated methods.

ashwaniarya-adyen: Done [here](https://github.com/Adyen/adyen-php-api-library/blob/82a321689e830721577e8566b58ee8c16a831c9c/templates-v7/api.mustache#L172)

src/Adyen/Model/BinLookup/Amount.php
ashwaniarya-adyen: Validation Removed

templates-v7/model_generic.mustache
ashwaniarya-adyen: Validation Removed

templates-v7/model_generic.mustache
ashwaniarya-adyen: Validation Removed

src/Adyen/Service/BinLookup/BinLookupApi.php
ashwaniarya-adyen: Support for custom host removed [here](https://github.com/Adyen/adyen-php-api-library/blob/82a321689e830721577e8566b58ee8c16a831c9c/templates-v7/api.mustache#L766) 
BinkLookAPI - [Here](https://github.com/Adyen/adyen-php-api-library/blob/82a321689e830721577e8566b58ee8c16a831c9c/src/Adyen/Service/BinLookup/BinLookupApi.php#L457)

templates-v7/api.mustache
ashwaniarya-adyen: Support for custom host removed [here](https://github.com/Adyen/adyen-php-api-library/blob/82a321689e830721577e8566b58ee8c16a831c9c/templates-v7/api.mustache#L766)

src/Adyen/Service/BinLookup/BinLookupApi.php
ashwaniarya-adyen: BinkLookAPI - [Here](https://github.com/Adyen/adyen-php-api-library/blob/82a321689e830721577e8566b58ee8c16a831c9c/src/Adyen/Service/BinLookup/BinLookupApi.php#L796)

src/Adyen/BaseService.php
ashwaniarya-adyen: Added support to handle both API Key and Basic Authentication [here](https://github.com/Adyen/adyen-php-api-library/blob/82a321689e830721577e8566b58ee8c16a831c9c/src/Adyen/BaseService.php#L20)

Added test cases to test [here](https://github.com/Adyen/adyen-php-api-library/blob/82a321689e830721577e8566b58ee8c16a831c9c/tests/Unit/BaseServiceTest.php#L46)

templates-v7/model_generic.mustache
ashwaniarya-adyen: Removed all the validation from SDK

PR #922, opened 2026-07-23
chore: convert deprecations to error

jeandersonbc:
**Description**

This is a follow-up action to prevent use of deprecated methods in the codebase. It uses `phpunit` to convert warnings to errors. Another commit also fixes a deprecation on tests.

**Tested scenarios**

- CI runs across different PHP versions


andrej-gajduk-adyen:
Need to fix:

Method ReflectionMethod::setAccessible() is deprecated since 8.5, as it has no effect since PHP 8.1

Either remove from tests, or guard against it based on PHP version

andrej-gajduk-adyen (approved):
LGTM

PR #936, opened 2026-09-02
Feature Terminal API regional endpoint configuration

ashwaniarya-adyen:
**Description**

- Adds `Client::setTerminalApiRegion()` and ensures the resolved regional endpoint is written to `endpointTerminalCloud`, which is used by `/sync`, `/async`, and
- The change preserves `terminalApiCloudEndpoint` for compatibility, supports valid legacy `region` configuration, gives `terminalApiRegion` priority, and throws
     `AdyenException` for unsupported regions.

**Tested scenarios**
    - Regional Terminal API URLs for sync, async, and connected terminals.
     - Region set before or after live environment configuration.
     - Default live and test endpoints.
     - Invalid regions and legacy `region` compatibility.

[PHP Library - Support Region in Terminal API ](https://youtrack.is.adyen.com/issue/IEX-3834/PHP-Library-Support-Region-in-Terminal-API)


src/Adyen/Client.php
```diff
@@ -175,14 +175,41 @@ public function setEnvironment(string $environment, ?string $liveEndpointUrlPref
...
+    {
+        if (!array_key_exists($region, Region::TERMINAL_API_ENDPOINTS_MAPPING)) {
+            throw new AdyenException("TerminalAPI endpoint for $region is not supported yet");
+        }
+        $this->config->set('terminalApiRegion', $region);
+        $environment = $this->config->getEnvironment();
+        if ($environment!=null) {
+            $this->setCloudEndPoint($region, $environment);
+        }
+    }
```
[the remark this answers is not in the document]
ashwaniarya-adyen: Thanks, fixed

src/Adyen/Service/PosPayment.php
```diff
@@ -33,14 +32,13 @@ public function __construct(Client $client)
...
         $environment = $this->getClient()->getConfig()->get('environment');
 
-        if (isset($region) && $environment == Environment::LIVE) {
-            $this->getClient()->getConfig()->set(
-                'endpointTerminalCloud',
-                Region::TERMINAL_API_ENDPOINTS_MAPPING[$region]
-            );
+        if ($region !== null && $environment == Environment::LIVE) {
+            $this->getClient()->setTerminalApiRegion($region);
         }
```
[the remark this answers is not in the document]
ashwaniarya-adyen: Thanks, fixed

tests/Unit/RegionTest.php
```diff
@@ -71,4 +73,16 @@ public function testTerminalApiEndpointsMapping(): void
             "TERMINAL_API_ENDPOINTS_MAPPING should match the expected mappings."
         );
     }
+
```
gcatanese, line 78: This looks good, my only suggestion is to add a test that verifies that, given a specific region (i.e. `US`), the `endpointTerminalCloud` url is set correctly. This test should have been already present to be  honest, but we can take now the opportunity to increase test coverage.

ashwaniarya-adyen: Done 😃

gcatanese (dismissed):
Looks good!

src/Adyen/Client.php
```diff
@@ -179,10 +179,19 @@ public function setEnvironment(string $environment, ?string $liveEndpointUrlPref
         $this->config->set('terminalApiCloudEndpoint', $endpoint);
     }
 
+    public function setRegion(string $region): void
+    {
+        if (!array_key_exists($region, Region::TERMINAL_API_ENDPOINTS_MAPPING)) {
```
candemiralp: Hello @ashwaniarya-adyen, thanks for the PR. Small comment here.

This condition makes this patch (and the setter) explicit to Terminal API. If you are planning to use `region` for generic purposes eventually, an exception might be thrown here for non-terminal related missing mapping values. I think this guardrail is good in theory can live in `retrieveCloudEndpoint()` method only.

On the other hand, if the patch is specific to Terminal API only, the naming and the config value should be more explicit rather than setting generic `region` value.

tests/Unit/RegionTest.php
```diff
@@ -71,4 +75,33 @@ public function testTerminalApiEndpointsMapping(): void
...
+    /**
+     * @throws AdyenException
+     */
+    public function testSetsUnitedStatesTerminalCloudEndpoint(): void
+    {
+        $client = new Client();
+        $client->setRegion(Region::US);
+        $client->setEnvironment(Environment::LIVE);
+
+        new PosPayment($client);
```
candemiralp, line 88: Unused object, can be removed.
```suggestion
```

src/Adyen/Client.php
ashwaniarya-adyen: Thanks. I initially used `setTerminalApiRegion()` and a Terminal API-specific config value ([here](https://github.com/Adyen/adyen-php-api-library/pull/936/changes/3a8dc567b6df5bcd8bca19409db9d222779460a5)), but after yesterday’s discussion I simplified the implementation by keeping  `setRegion()` and handling the endpoint in `PosPayment`.  😄 
`region` was already used by `PosPayment` for Terminal API configuration, so reverting to it preserves the existing behavior and avoids requiring changes from users. What do you think ?

tests/Unit/RegionTest.php
ashwaniarya-adyen: The PosPayment instance is intentionally created because its constructor applies the regional Terminal API endpoint. The object itself is not used afterward, but removing it would no longer test the current behavior.

candemiralp: Thank you @ashwaniarya-adyen, that makes sense!

src/Adyen/Client.php
ashwaniarya-adyen: Also yes, it makes sense if region is used for another purpose later, validating it against Terminal API regions here could cause an incorrect error. It would be better to validate the region only when fetching the Terminal API URL. Thanks.

candemiralp (approved):

src/Adyen/Client.php
candemiralp: Thank you @ashwaniarya-adyen!

gcatanese (approved):

PR #937, opened 2026-09-14
Checkout openapi generator upgrade

ashwaniarya-adyen:
Checkout openapi generator upgrade


templates-v7/api.mustache
```diff
@@ -169,9 +171,9 @@ use {{modelPackage}}\ObjectSerializer;
...
+    public function {{operationId}}({{^exts.x-group-parameters}}{{#pathParams}}{{dataType}} ${{paramName}}, {{/pathParams}}{{#bodyParams}}{{dataType}} ${{paramName}}, {{/bodyParams}}{{#formParams}}{{#required}}{{#notRequiredOrIsNullable}}?{{/notRequiredOrIsNullable}}{{dataType}} ${{paramName}}, {{/required}}{{/formParams}}{{#queryParams}}{{#required}}{{#notRequiredOrIsNullable}}?{{/notRequiredOrIsNullable}}{{dataType}} ${{paramName}}, {{/required}}{{/queryParams}}{{#formParams}}{{^required}}?{{dataType}} ${{paramName}} = {{#defaultValue}}{{{.}}}{{/defaultValue}}{{^defaultValue}}null{{/defaultValue}}, {{/required}}{{/formParams}}{{#queryParams}}{{^required}}?{{dataType}} ${{paramName}} = {{#defaultValue}}{{{.}}}{{/defaultValue}}{{^defaultValue}}null{{/defaultValue}}, {{/required}}{{/queryParams}}{{#servers}}{{#-first}}?int $hostIndex = null, array $variables = [], {{/-first}}{{/servers}}?\{{invokerPackage}}\RequestOptions $requestOptions = null{{/exts.x-group-parameters}}{{#exts.x-group-parameters}}$associative_array{{/exts.x-group-parameters}}): {{#returnType}}{{.}}{{/returnType}}{{^returnType}}void{{/returnType}}
```
ashwaniarya-adyen, line 174: Earlier, parameters were generated in the same order as the API specification. This exposed the optional  headers as a PHP method argument before the required request body, causing PHP 8.4 warnings and making it behave as required. The template now excludes header parameters from method arguments, places required parameters before optional ones, and handles headers through `RequestOptions`

templates-v7/model_generic.mustache
```diff
@@ -241,7 +241,7 @@ class {{classname}} {{#parentSchema}}extends {{{parent}}}{{/parentSchema}}{{^par
 
         {{/parentSchema}}
         {{#vars}}
-        $this->setIfExists('{{name}}', $data ?? [], {{#defaultValue}}{{{defaultValue}}}{{/defaultValue}}{{^defaultValue}}null{{/defaultValue}});
```
ashwaniarya-adyen, line 244: The initial v7 generation automatically assigned default values from the API specification when creating a model. This caused fields that the user did not provide, to be included in outgoing requests. The previous Checkout version left unset fields as `null` and omitted them during serialization. This change restores that behavior.

tests/Unit/ModelBasedCheckoutTest.php
```diff
@@ -21,21 +21,22 @@
...
+     * @throws AdyenException
      */
     public function testPaymentMethodsSuccess($jsonFile, $httpStatus)
     {
         // create Checkout client
-        $client = $this->createMockClient($jsonFile, $httpStatus);
-
-        // initialize service
-        $service = new \Adyen\Service\Checkout\PaymentsApi($client);
+        $client = $this->createMockSerializerClient($jsonFile, $httpStatus);
```
ashwaniarya-adyen: These tests previously created generated Checkout APIs using the legacy `Adyen\Client`. The v7-generated APIs now require an `Adyen\Configuration` and a Guzzle HTTP client separately. The tests now extend `BaseTest` and use its existing helpers to create both objects.

tests/Unit/BaseTest.php
```diff
@@ -13,13 +13,13 @@
...
-     * @param string $jsonFile
+     * @param string|null $jsonFile
      * @param int $httpStatus
      * @return Client
      */
-    protected function createMockSerializerClient(string $jsonFile, int $httpStatus, &$container = []): Client
+    protected function createMockSerializerClient(?string $jsonFile, int $httpStatus, &$container = []): Client
     {
-        $json = file_get_contents($jsonFile);
+        $json = $jsonFile !== null ? file_get_contents($jsonFile) : '';
```
ashwaniarya-adyen, line 22: This test helper previously required a JSON file. Some API calls return no response body, such as a successful delete request with status 204. The helper now allows no JSON file and uses an empty response instead.

tests/Unit/ModelBasedCheckoutTest.php
```diff
@@ -292,14 +301,15 @@ public static function successGetStoredPaymentMethodsProvider()
...
     {
-        $client = $this->createMockClient(null, 204);
+        $client = $this->createMockSerializerClient(null, 204);
+        $config = $this->createConfiguration();
+        $service = new RecurringApi($config, $client);
 
-        $service = new RecurringApi($client);
-
-        $service->deleteTokenForStoredPaymentDetails("123");
+        $service->deleteTokenForStoredPaymentDetails("paymentMethodId", "shopperReference", "merchantAccount");
```
ashwaniarya-adyen: This test previously passed only the stored payment method ID. In Checkout v72, the shopper reference and merchant account are also required, so the test now provides all three values.

tests/Unit/ModelBasedCheckoutTest.php
```diff
@@ -53,41 +54,34 @@ public static function successPaymentMethodsProvider()
...
-            return ObjectSerializer::sanitizeForSerialization($result);
-        };
-        // second function calling to json encode + decode
-        $func2 = function () use ($result) {
-            return json_decode(json_encode($result->jsonSerialize()), true);
-        };
-        // And assert that the result is equal to a deep json encode/decode
-        #$this->assertEquals($result->toArray(), json_decode(json_encode($result->jsonSerialize()), true));
-        $this->assertEquals(ObjectSerializer::sanitizeForSerialization($result), json_decode(json_encode($result->jsonSerialize()), true));
+        $this->assertEquals(
```
ashwaniarya-adyen: This test previously compared an object with an array, so it failed even when both contained the same data. It now compares two arrays and correctly checks that `toArray()` returns the same data as JSON serialization.

templates-v7/model_generic.mustache
```diff
@@ -241,12 +241,14 @@ class {{classname}} {{#parentSchema}}extends {{{parent}}}{{/parentSchema}}{{^par
...
+        $this->setIfExists('{{name}}', $data ?? [], null);
         {{/vars}}
         {{#discriminator}}
 
         // Initialize discriminator property with the model name.
-        $this->container['{{discriminatorName}}'] = static::$openAPIModelName;
+        if (!array_key_exists('{{discriminatorName}}', $data ?? [])) {
```
ashwaniarya-adyen, line 249: The first v7 generation always replaced the model discriminator after loading constructor data. For example, a supplied payment method type of `paypal` was replaced by `CheckoutPaymentMethod`. The generated model name is now used only when the input does not already contain a type.

src/Adyen/Configuration.php
```diff
@@ -120,6 +119,20 @@ class Configuration
      */
     protected ?string $liveEndpointUrlPrefix = null;
 
+    /**
+     * Adyen library name, sent with every request as the adyen-library-name header
+     *
+     * @var string
+     */
+    protected string $libraryName = self::LIB_NAME;
```
ashwaniarya-adyen, line 127: The initial v7 configuration lost this information, so this PR adds the library name and version to `Configuration`.

src/Adyen/Configuration.php
```diff
@@ -300,24 +313,17 @@ public function getPassword(): string
...
-     * Gets the user agent of the api client
+     * Gets the User-Agent for HTTP requests: the application name (if set) followed by
+     * the library name and version, e.g. "MyShopApp adyen-php-api-library/30.0.2"
      *
      * @return string user agent
      */
     public function getUserAgent(): string
     {
-        return $this->userAgent;
+        $suffix = self::LIB_NAME . '/' . self::LIB_VERSION;
```
ashwaniarya-adyen, line 344: The initial v7 generation used `OpenAPI-Generator/1.0.0/PHP` as the User-Agent. This PR restores the SDK identity:
    adyen-php-api-library/{version}
If an application name is configured, it is placed before the library identity.

templates-v7/api.mustache
```diff
@@ -670,9 +669,13 @@ use {{modelPackage}}\ObjectSerializer;
             $requestOptions
         );
 
+        $headers['adyen-library-name'] = $this->config->getLibraryName();
```
ashwaniarya-adyen, line 672: Compared with the initial generation, every request now includes `adyen-library-name` and `adyen-library-version`. This restores the identification headers used by the Adyen PHP library.

src/Adyen/BaseService.php
```diff
@@ -79,4 +83,62 @@ public function createBaseUrl(string $url): string
...
+     * Adds or overwrites the applicationInfo adyenLibrary name and version on a request model
+     * and merges the adyenPaymentSource, externalPlatform and merchantApplication values
+     * configured on the Configuration, mirroring the behaviour of the array based services.
+     * Request models without an applicationInfo field are returned untouched and
+     * merchant-provided values are kept unless a matching value is configured.
+     *
+     * @param object|null $requestModel
+     * @return object|null
+     */
+    protected function injectApplicationInfo(?object $requestModel): ?object
```
ashwaniarya-adyen, line 97: Restored the previous behavour. The initial v7 generation dropped it. The `injectApplicationInfo()` helper restores the old behavior, it handles `applicationInfo` set as an array or a model object and leaves any other merchant-provided fields untouched - [here](https://github.com/Adyen/adyen-php-api-library/blob/125dc2e9666a3d765068fc33116f0ba1e28288cb/src/Adyen/Service/AbstractResource.php#L132)

templates-v7/api.mustache
```diff
@@ -670,9 +669,13 @@ use {{modelPackage}}\ObjectSerializer;
...
+        $headers['adyen-library-version'] = $this->config->getLibraryVersion();
+
         // for model (json/xml)
         {{#bodyParams}}
         if (isset(${{paramName}})) {
+            ${{paramName}} = $this->injectApplicationInfo(${{paramName}});
```
ashwaniarya-adyen, line 678: The request model is passed through `injectApplicationInfo()` before it is serialized. This ensures supported request bodies contain `applicationInfo.adyenLibrary` without adding the same logic manually to every generated operation.

src/Adyen/Configuration.php
```diff
@@ -436,9 +463,17 @@ public function getKeyFile(): ?string
      *
      * @param string $environment
      * @return $this
+     * @throws AdyenException
      */
     public function setEnvironment(string $environment): self
     {
+        if (!in_array($environment, [Environment::TEST, Environment::LIVE], true)) {
```
ashwaniarya-adyen, line 470: The initial v7 changes accepted any environment string. This PR changes to only `test` and `live` are accepted. This prevents unsupported values.

.github/workflows/release.yml
```diff
@@ -33,7 +33,7 @@ jobs:
         with:
           token: ${{ secrets.ADYEN_AUTOMATION_BOT_ACCESS_TOKEN }}
           develop-branch: main
-          version-files: src/Adyen/Client.php README.md
+          version-files: src/Adyen/Client.php src/Adyen/Configuration.php README.md
```
ashwaniarya-adyen, line 36: The release workflow now updates both `Client.php` and `Configuration.php` when creating a new release, until we removed Client.php completely

tests/Unit/BaseServiceTest.php
```diff
@@ -238,4 +241,79 @@ public function testCreateBaseUrlLiveOtherEndpoint()
         $expected = 'https://kyc-live.adyen.com/lem/v3/legalEntities';
         $this->assertEquals($expected, $service->createBaseUrl($url));
     }
+
+    /**
+     * @covers \Adyen\BaseService::injectApplicationInfo
+     */
+    public function testInjectApplicationInfo()
```
ashwaniarya-adyen, line 248: Added test cases for injecting ApplicationInfo

src/Adyen/Configuration.php
```diff
@@ -120,6 +119,41 @@ class Configuration
...
+     * @var string
+     */
+    protected string $libraryVersion = self::LIB_VERSION;
+
+    /**
+     * Adyen payment source name and version, included in applicationInfo.adyenPaymentSource of requests
+     *
+     * @var array{name: string, version: string}|null
+     */
+    protected ?array $adyenPaymentSource = null;
```
ashwaniarya-adyen, line 141: Earlier ,  `adyenPaymentSource`, `externalPlatform`, and `merchantApplication` were stored in the client configuration and added to `applicationInfo`, added them in Configuration

templates-v7/model_generic.mustache
```diff
@@ -303,7 +305,7 @@ class {{classname}} {{#parentSchema}}extends {{{parent}}}{{/parentSchema}}{{^par
         } else {
             $nullablesSetToNull = $this->getOpenAPINullablesSetToNull();
             $index = array_search('{{name}}', $nullablesSetToNull);
-            if ($index !== FALSE) {
+            if ($index !== false) {
```
ashwaniarya-adyen, line 308: Changes `FALSE` to lowercase `false` to follow the project PHP coding standard. This is formatting only and does not change the behavior.

ashwaniarya-adyen:
Previously, most Checkout operations were tested in one large `CheckoutTest`. Split this into one test file for each generated API: Payments, Modifications, Orders, Donations, Payment Links, Recurring, and Utility. This makes it easier to find a test failure and update the affected API after a future regeneration.

Kept model-specific behavior in `ModelBasedCheckoutTest`. That file focuses on array-based construction, nested models, nullable fields, serialization, and discriminator values. The API test files focus on sending requests and reading responses.

The common setup is kept in `BaseTest`. It creates the `Configuration`, the mocked Guzzle client, and the HTTP response. It also supports operations that return HTTP 204 with no response body.

Where it makes sense for an operation, test three common ways of using the generated models:
    - create the request with model setters and read the response as a model;
    - create the request model from an array;
    - convert the response to an array with `toArray()`.

OpenAPI Generator 7 also creates several versions of each API operation. Testing every version for every endpoint would repeat the same generated template logic, so used representative operations:
       - `payments()` covers the normal synchronous call;                                                                                                                                                                                                                                                                   
       - `paymentsWithHttpInfo()` verifies the response model, HTTP status, and response headers;                                                                                                                                                                                                                           
       - `paymentsAsync()` verifies that the promise resolves to the expected model;                                                                                                                                                                                                                                        
       - `paymentsAsyncWithHttpInfo()` verifies the model, status, headers, and that `RequestOptions` such as the idempotency key are applied;                                                                                                                                                                              
       - `deleteTokenForStoredPaymentDetailsWithHttpInfo()` verifies a `void` operation returning HTTP 204;    

The async error tests are present but currently skipped because error handling is being addressed separately in a different ticket.

src/Adyen/Model/Checkout/AccountInfo.php
```diff
@@ -96,24 +93,24 @@ class AccountInfo implements ModelInterface, ArrayAccess, \JsonSerializable
...
       * @var boolean[]
       */
-    protected static $openAPINullables = [
+    protected static array $openAPINullables = [
         'accountAgeIndicator' => false,
         'accountChangeDate' => false,
         'accountChangeIndicator' => false,
         'accountCreationDate' => false,
         'accountType' => false,
-        'addCardAttemptsDay' => true,
```
gcatanese, line 105: Looks like several attributes are now marked as not nullable (they were nullable), it looks correct as they are now marked as nullable in the OpenAPI spec. Just double checking with you I understand correctly? @ashwaniarya-adyen

ashwaniarya-adyen: Yes, they were marked as non nullable in OpenAPI spec, but to be sure I checked in version 71, they were marked as non nullable as well. So to be sure, I asked droid whats the reason, and it mentions -
`They were only marked nullable in the old generated code because of a template workaround that forced all integer fields to be treated as nullable`
[Here](https://github.com/Adyen/adyen-php-api-library/blob/125dc2e9666a3d765068fc33116f0ba1e28288cb/templates/model_generic.mustache#L40) is the old template

